PR #2, opened 2016-04-04
Initial work on fuzzy matching

lakinwecker:
This uses levenshtein distance of the strings and some regexes to make the commands more resilient. It also uses the channel name to determine which tournament is appropriate.

Only the mods command has been improved in this way, but the facilities should be easily ported to other commands.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -1,0 +1,44 @@
+# -*- mode: ruby -*-
+# vi: set ft=ruby :
+
+require 'json'
+
+# All Vagrant configuration is done below. The "2" in Vagrant.configure
+# configures the configuration version (we support older styles for
+# backwards compatibility). Please don't change it unless you know what
+# you're doing.
+Vagrant.configure(2) do |config|
+  config.vm.box = "ubuntu/trusty64"
+
+  # Disable automatic box update checking. If you disable this, then
+  # boxes will only be checked for updates when the user runs
+  # `vagrant box outdated`. This is not recommended.
+  # config.vm.box_check_update = false
+
+  # Create a private network, which allows host-only access to the machine
+  # using a specific IP.
+  # config.vm.network "private_network", ip: "192.168.33.10"
+
+  # Create a public network, which generally matched to bridged network.
+  # Bridged networks make the machine appear as another physical device on
+  # your network.
+  # config.vm.network "public_network"
+
+  # Provider-specific configuration so you can fine-tune various
+  # backing providers for Vagrant. These expose provider-specific options.
+  # View the documentation for the provider you are using for more
+  # information on available options.
+  config.vm.provider "virtualbox" do |vb|
+     # Display the VirtualBox GUI when booting the machine
+     # vb.gui = true
+
+     # Customize the amount of memory on the VM:
+     vb.memory = 2048
+  end
+
+  config.vm.hostname = "modster"
+
+
+  config.vm.provision :shell, :path => "./vagrant-dependencies.sh"
+
+end
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -1,8 +1,6 @@
 # -*- mode: ruby -*-
 # vi: set ft=ruby :
 
-require 'json'
-
 # All Vagrant configuration is done below. The "2" in Vagrant.configure
 # configures the configuration version (we support older styles for
 # backwards compatibility). Please don't change it unless you know what
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -36,7 +36,7 @@ Vagrant.configure(2) do |config|
      vb.memory = 2048
   end
 
-  config.vm.hostname = "modster"
+  config.vm.hostname = "chesster"
 
 
   config.vm.provision :shell, :path => "./vagrant-dependencies.sh"
--- a/chesster.js
+++ b/chesster.js
@@ -2,6 +2,7 @@ var async = require("async");
 var Botkit = require('botkit');
 var GoogleSpreadsheet = require("google-spreadsheet");
 var fs = require('fs');
+var levenshtein = require('fast-levenshtein');
 
 var SHEET_URL = "https://lichess4545.slack.com/files/mrlegilimens/F0VNACY64/lichess4545season3-graphs";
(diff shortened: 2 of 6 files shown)


Vagrantfile
```diff
@@ -0,0 +1,44 @@
+# -*- mode: ruby -*-
```
endrawes0: I am not familiar with Vagrant. Can you give me some background? I would love to learn a little.

chesster.js
```diff
@@ -26,6 +27,34 @@ var BOARD_5_RATING = 16;
...
 
+var LONEWOLF_CHANNEL_NAMES = [
+    "lonewolf-general",
+    "lonewolf-gamelinks",
+    "lonewolf-schedule",
+    "lonewolf-results",
+    "unstable_bot-lonewolf"
+];
+var TEAM_CHANNEL_NAMES = [
+    "general",
```
endrawes0: this can include team-general as well

chesster.js
```diff
@@ -51,6 +80,95 @@ var channels = {
         return "<#"+this.getId(name)+">";
     }
 };
+var channelsByID = { };
+
+
+var fuzzy_match = function (search_string, targets) {
+    if (!targets) { return null; } // TODO: not sure this is the right choice ...
```
endrawes0: I would prefer to throw here. errors should be caught and logged in the handler.

chesster.js
```diff
@@ -51,6 +80,95 @@ var channels = {
...
+        var distance = item[0];
+        var match = item[1];
+        if (distance == min_distance) {
+            choices.push(match);
+        }
+    });
+
+    return choices;
+}
+var get_command_and_targets = function(message, commands, arg_string) {
```
endrawes0: I dont mind defining functions this way - as far as I know, it is effectively the same, just a style difference - but I would like to be consistent.

Can we make this consistent with other function definition style? or make the others the same as this. Either way is fine with me.

chesster.js
```diff
@@ -51,6 +80,95 @@ var channels = {
...
+    var target = "general";
+    var channel = channelsByID[message.channel];
+    if (channel && channel.name) {
+        target = channel.name;
+        console.log("Setting target to: ", channel.name);
+    }
+    var args = arg_string;
+    if (args) {
+        args = args.split(" ");
+        if (args.length > 2) {
```
endrawes0: again - I would prefer to throw when I encounter an error. let me know if this is not possible for one reason or another.

chesster.js
```diff
@@ -322,36 +441,36 @@ function sayLoneWolfMods(convo){
...
-],function(bot,message) {
-    exception_handler(bot, message, function(){
-        bot.reply(message, prepareSummonLoneWolfModsMessage());
-    });
-});
-
-controller.hears([
-    'mods summon'
-],[
+    "^mods$",
```
endrawes0: the regex should be useful. I'll consider apply it elsewhere too. Thanks.

chesster.js
```diff
@@ -610,6 +729,7 @@ function exception_handler(bot, message, todo){
             "\nDatetime: " + new Date() + 
             "\nMessage: " +  JSON.stringify(message) + 
             "\nError: " + JSON.stringify(e) + 
+            "\nStack: " + e.stack +
```
endrawes0: maybe I have not been paying attention, but didnt the stringify on the error already show the stack? 

If not, good call. Thanks.

endrawes0:
Laikin - I made some comments on your request. Can you take a look and reply?

Thanks!

Vagrantfile
lakinwecker: https://www.vagrantup.com/  it's an easy way to make reproduceable lightweight VMs.  In this case it provisions a server based on ubuntu 14.04.  Then it runs the vagrant-dependencies script on it which sets up the environment.

If you install the latest version of vagrant, and in the directory with the Vagrantfile, you type: ```vagrant up``` then it will download and install this machine.  It will also start it running.  ```vagrant ssh``` will connect you to that machine, then you can do cd chesster and ./start_debug and it should run on that machine.  ```vagrant halt``` will bring the machine down. One minor annoying thing about vagrant is that the machines stay running in the background until you stop them. :/

Eithe way, I should clean this up - it still refers to "modster". It should refer to chesster.  I'll do that now.

lakinwecker: Also, it maps your source from that local directory to a directory called /vagrant/ within the virtual machine. I actually symlink it to /home/vagrant/chesster

chesster.js
lakinwecker: Good call.

chesster.js
lakinwecker: It was not for me.  I'll test it again.

lakinwecker: Confirmed. It was not showing the stack trace for me.

lakinwecker:
I've addressed all of the items you mentioned. 

I'd also like to point out the change to ./start and ./start_debug by using source of ~/key.sh instead of relying on my own memory to avoid committing and pushing a key.  I'm not super happy with the approach I took, but I would prefer that we use a method that allows me to save the key and use it without having to worry about remember to remove it before I commit.  Thoughts?

lakinwecker:
Also, I'm not particularly happy with the function names "fuzzy_match" and "get_command_and_target", although I couldn't think of anything better at the time, any suggestions?

I was thinking the pattern could be re-used for other commands.

chesster.js
endrawes0: OK. Then thanks for adding it!

Vagrantfile
endrawes0: Thanks for all the info. Very cool. Ill have to investigate further.

endrawes0:
Thanks for helping out. I will definitely want to apply the distance function to other chesster commands.

We should look into modding the bot so we can apply the distance function to the pattern matching args themselves. That would be cool.

Re: Keys
This approach is ok for now. I generally just dont add those files to my commit to avoid that issue. but whatever you need is ok. If I come up with something better, ill let you know. One thing that comes to mind is making the key an argument to the script and then creating scripts that call into the start scripts. We can check in the start scripts and leave the wrappers out of the git.

Re: Function Names
They are ok for what they do. Maybe some refactoring will help you name it more clearly.

PR #3, opened 2016-04-09
Refresh

endrawes0:


chesster.js
```diff
@@ -27,177 +30,161 @@ var BOARD_5_RATING = 16;
...
+
+/* static entry point */
+
+var config;
+
+exception_handler(function(){ 
+    var config_file = process.argv[2] || "config.json";
+
+    console.log("Loading configuration from " + config_file);
+    config = JSON.parse(fs.readFileSync(config_file, 'utf8'));
```
lakinwecker: the encoding is technically utf-8 (with the dash). I think most contexts will happily accept the non-dash version, but some don't.

chesster.js
```diff
@@ -27,177 +30,161 @@ var BOARD_5_RATING = 16;
...
-        retval.target = "lonewolf";
-    }
-    return retval;
-}
-
-controller.spawn({
-  token: process.env.token
-}).startRTM(function(err, bot) {
+function update_users(bot){
+    users.list = [];
```
lakinwecker: As this is async, there is a period of time during which the bot.api.users.list function below has not finished, but we have an empty users.list and functions which depend upon it will fail. I think this line should be moved into the callback below.  Basically, if we have finished AND there are no errors, then we reset the list.

chesster.js
```diff
@@ -27,177 +30,161 @@ var BOARD_5_RATING = 16;
...
-                users[member.name] = member;
+                users.list[member.name] = member;
             }
         }
         console.log("info: got users");
     });
+}
 
+function update_channels(bot){
+    channels.list = [];
```
lakinwecker: Same goes here.

chesster.js
```diff
@@ -27,177 +30,161 @@ var BOARD_5_RATING = 16;
...
         }
         console.log("info: got channels");
     });
+
+}
+
+/* 
+   updates the user list
+   updates the channel lists
+   then repeats in new threat
```
lakinwecker: threat - > thread

changelog
```diff
@@ -11,3 +11,10 @@ April 3, 2015:
 3. Fixed for 26 teams
 4. Added board commands
 5. Fixed LoneWolf/Team Typo in Channel Details
+
+April 9, 2015:
+1. Fixed Worksheet to be based on name
+2. Added refresh for users and channels
+3. Refactored fuzzy matching
+4. Updated to JSON configuration files
+5. ADded PING for RTM
```
lakinwecker: ADded - Added

chesster.js
```diff
@@ -27,177 +30,161 @@ var BOARD_5_RATING = 16;
...
+
+controller.spawn({
+  token: config.token
+}).startRTM(function(err, bot) {
+
     if (err) {
         throw new Error(err);
     }
+    
+    //rrefresh your user and channel list every 2 minutes
```
lakinwecker: ```// refresh the user and channel list every 2 minutes```

chesster.js
```diff
@@ -27,177 +30,161 @@ var BOARD_5_RATING = 16;
...
+    //woudl be nice if this was push model, not poll but oh well.
```
lakinwecker: ```// would be nice if this was a push model, not poll but oh well. ```

chesster.js
endrawes0: Good call
On Apr 10, 2016 11:43 AM, lakinwecker <notifications@github.com> wrote:In chesster.js:
> -    };
> -    if (team_votes == wolf_votes) {
> -        console.log("Ambiguous target!");
> -    } else if (team_votes > wolf_votes) {
> -        retval.target = "team";
> -    } else if (wolf_votes > team_votes) {
> -        retval.target = "lonewolf";
> -    }
> -    return retval;
> -}
> -
> -controller.spawn({
> -  token: process.env.token
> -}).startRTM(function(err, bot) {
> +function update_users(bot){
> +    users.list = [];

As this is async, there is a period of time during which the bot.api.users.list function below has not finished, but we have an empty users.list and functions which depend upon it will fail. I think this line should be moved into the callback below.  Basically, if we have finished AND there are no errors, then we reset the list.

chesster.js
endrawes0: Ha. Must have been playing chess when I wrote that
On Apr 10, 2016 11:44 AM, lakinwecker <notifications@github.com> wrote:In chesster.js:
>              }
>          }
>          console.log("info: got channels");
>      });
> +
> +}
> +
> +/* 
> +   updates the user list
> +   updates the channel lists
> +   then repeats in new threat

threat - > thread

lakinwecker:
Aside from some minor issues, this looks good to go.

endrawes0:
Great. I'm make the changes, update the request and you can pull.
On Apr 10, 2016 11:53 AM, lakinwecker <notifications@github.com> wrote:Aside from some minor issues, this looks good to go.

chesster.js
endrawes0: https://github.com/nodejs/node/blob/master/lib/buffer.js#L299

If you are curious. I wanted to know for certain so I went hunting.

PR #7, opened 2016-04-19
Switching to use the cell base API to avoid munching yummy formulas.

lakinwecker:


spreadsheets.js
```diff
@@ -206,38 +214,93 @@ function find_pairing(service_account_auth, spreadsheet_key, white, black, callb
...
-                        potential_rows.push([row, true]);
-                    }
-                });
-                // Only update it if we found an exact match
-                if (potential_rows.length > 1) {
-                    return callback("Unable to find pairing. More than one row was returned!");
-                } else if(potential_rows.length < 1) {
-                    return callback("Unable to find pairing. No rows were returned!");
+            options = _.clone(options);
+            var defaults = {
```
endrawes0: I have some constants that I intend to add to the config. The could be used here to make this more flexible.

I'll get to that later. This is fine for the rush we are in to get Chesster back up.

lakinwecker: :+1:  to constants as defaults and putting them into the config

spreadsheets.js
```diff
@@ -206,38 +214,93 @@ function find_pairing(service_account_auth, spreadsheet_key, white, black, callb
...
+        ) {
+            return true;
+        }
+        return false;
+    }
+    get_rows(
+        service_account_auth,
+        spreadsheet_key,
+        options,
+        function(err, rows) {
```
endrawes0: I wonder if it makes sense to diverge from the interface defined by the looks API and include a filter functor as a parameter. Then it might be easier to extend this to some of our other applications.

endrawes0: Previous* API

endrawes0:
I will have an opportunity to do a bunch of coding this weekend.

Until then, I think this looks great.

Nothing major that I can see from my cell phone.

spreadsheets.js
lakinwecker: I actually originally implemented this as a filter functor passed as a parameter. But in the end switched back to this because underscore make it so easy to just add like this.  It's so easy that I thought that this was a more general API and probably more useful in other cases.  It's a simple change either way, so I'm happy to move back to that or see it moved back to that.

PR #9, opened 2016-04-20
A set of very simple fixes, including fixing the colon issue that @endrawes0 identified

lakinwecker:
1. Fixing the tests to run outside of that round (properly using the extrema.reference_date).
2. Removing an unused test method.
3. Fixing the list of dependencies in the README.md
4. Fixing the parsing of dates that had the colon in them.

### DIFF
--- a/chesster.js
+++ b/chesster.js
@@ -1263,7 +1263,6 @@ controller.on('ambient', function(bot, message) {
                 scheduling_reply_cant_schedule_others(bot, message);
                 return;
             }
-            console.log(message);
             results.white = white.name;
             results.black = black.name;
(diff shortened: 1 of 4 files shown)


chesster.js
```diff
@@ -1263,7 +1263,6 @@ controller.on('ambient', function(bot, message) {
                 scheduling_reply_cant_schedule_others(bot, message);
                 return;
             }
-            console.log(message);
```
endrawes0: That brings up a good question, I see this is being removed. But can we catch standard out and write to another log? 

I'll probably add this to the start script

PR #11, opened 2016-04-23
Attempt to quiet @chesster in the scheduling channels.

lakinwecker:


spreadsheets.js
```diff
@@ -9,17 +9,94 @@ var EXTREMA_DEFAULTS = {
...
+    "HH:mm",
+    "H:mm",
+    "H:m"
+];
 var DATE_FORMATS = [
-    "YYYYMMDD HHmm",
-    "HHmm DDMMYYYY"
 ];
+BASE_DATE_FORMATS.forEach(function(date_format) {
+    BASE_TIME_FORMATS.forEach(function(time_format) {
```
endrawes0: Nice use of static data.

spreadsheets.js
```diff
@@ -9,17 +9,94 @@ var EXTREMA_DEFAULTS = {
...
+        var time_format2 = time_format.replace(/[:]/g, '');
+        DATE_FORMATS.push(date_format + " " + time_format);
+        DATE_FORMATS.push(time_format + " " + date_format);
+        DATE_FORMATS.push(date_format2 + " " + time_format2);
+        DATE_FORMATS.push(time_format2 + " " + date_format2);
+        DATE_FORMATS.push(date_format2 + " " + time_format);
+        DATE_FORMATS.push(time_format + " " + date_format2);
+        DATE_FORMATS.push(date_format + " " + time_format2);
+        DATE_FORMATS.push(time_format2 + " " + date_format);
+    });
```
endrawes0: This structure just got much bigger

lakinwecker: Yeah, it's huge and slow.  Which is why I need to short circuit the date parsing below.

endrawes0: Yes, i saw that. Not much you can do though. Unless we can order these in some way, brute force is out only option.

endrawes0:
Looks fine. Pulling.

PR #12, opened 2016-04-24
Result parsing

endrawes0:


chesster.js
```diff
@@ -1329,3 +1322,71 @@ controller.on('ambient', function(bot, message) {
...
+            throw e;
+        }
+    });
+});
+
+function reply_permission_failure(bot, message){
+    bot.reply(message, "Sorry, you do not have permissin to update that pairing.");
+}
+
+function resultereply_missing_pairing(bot, message){
```
lakinwecker: This function is missing an _ in it's name.

spreadsheets.js
```diff
@@ -126,6 +126,22 @@ var WORDS_TO_IGNORE = [
...
+var VALID_RESULTS = {
+    "0-0":"0-0",
+    "1-0":"1-0",
+    "0-1":"0-1",
+    "1/2-1/2":"1/2-1/2",
+    "0.5-0.5":"1/2-1/2",
+    "1X-0F":"1X-0F",
+    "0F-1X":"0F-1X",
+    "0F-0F":"0F-0F",
+    "1/2X-1/2X":"1/2X-1/2X",
```
lakinwecker: theino pointed out that it is 1/2Z - 1/2Z

spreadsheets.js
```diff
@@ -422,8 +438,103 @@ function update_schedule(service_account_auth, key, colname, format, schedule, c
...
+    //assuming we found 2 tokens, we should convert them to player names
+    if(player_tokens.length == 2){
+        players.white = player_tokens[0];
+        players.black = player_tokens[1];
+    }
+
+    return players;
+}
+
+function filter_player_tokens(tokens){
```
lakinwecker: This, I like this.   I had been considering doing something similar, but never got around to looking for the documentation on which characters a slack username ID might contain. I just tried looking it up again to see if I could verify this regex, but couldn't find it. 

Did you find documentation on which characters the usernames can contain?

spreadsheets.js
```diff
@@ -422,8 +438,103 @@ function update_schedule(service_account_auth, key, colname, format, schedule, c
...
+        service_account_auth, 
+        key, 
+        white.name, 
+        black.name, 
+        function(err, row, reversed){
+            if(err){
+                return callback(err);
+            }
+            var result_cell = row[colname];
+            console.log(result_cell);
```
lakinwecker: Did you intentionally leave this in? I am ok with us leaving in a certain amount of logging, and think it would be worth using the various console.* methods, like console.error for actual errors, console.log for just logging info etc.

spreadsheets.js
```diff
@@ -422,8 +438,103 @@ function update_schedule(service_account_auth, key, colname, format, schedule, c
...
+               var lhs = result.split('-')[0];
+               var rhs = result.split('-')[1];
+               result = rhs + '-' + lhs;
+            }
+
+            var gamelink = ' ';            
+
+            var formula = result_cell.formula;
+            if(formula && formula.toUpperCase().includes("HYPERLINK")){
+               gamelink = formula.split("\"")[1];
```
lakinwecker: This doesn't look quite right to me. When I split:
```
=HYPERLINK(\"\", \"<link>\")".split("\"");
["=HYPERLINK(", "", ", ", "<link>", ")"]
```
The link is element number 3

spreadsheets.js
```diff
@@ -422,8 +438,103 @@ function update_schedule(service_account_auth, key, colname, format, schedule, c
...
+            }
+
+            result_cell.formula = '=HYPERLINK("' + gamelink + '","' + result + '")';
```
lakinwecker: I've been pining for python's str.format or cppformat style formatting, and npm has it:

https://www.npmjs.com/package/string-format

you don't need to change this pull request, but moving forward, I think I will switch to using this so we can do:

```format('=HYPERLINK("{}", "{}")', result, game_link);```

File: spreadsheets.js (hunk 5 above)
Comment (lakinwecker) [inline spreadsheets.js]: Nevermind. I realized that the HYPERLINK example you gave me in chat has the order of the link/text switched. you are good to go.

File: test/test_spreadsheets.js [hunk 7]
```diff
@@ -29,17 +29,17 @@ describe('scheduling', function() {
             options.extrema.iso_weekday = 1;
             options.extrema.reference_date = moment.utc("2016-04-07");
             var bounds = spreadsheets.get_round_extrema(options);
-            assert.equal(bounds.start.format(), "2016-04-04T11:00:00+00:00")
-            assert.equal(bounds.end.format(), "2016-04-11T11:00:00+00:00")
+            assert.equal(bounds.start.format(), "2016-04-04T11:00:00Z")
```
lakinwecker: I'm actually quite surprised by this change. These tests were passing on my system.  I wonder if the `.format` method is locale specific and we have different locales?  if so, we may want to consider passing in a specific format to each of these calls so that the locale doesn't change the test results.  

Again, don't worry about changing this pull request, Once this is merged I will pull and look into it.

spreadsheets.js
```diff
@@ -422,8 +438,103 @@ function update_schedule(service_account_auth, key, colname, format, schedule, c
     });
 }
 
+// parse the input string for a results update
+function parse_result(input_string){
```
lakinwecker: I like the way you organized this code. if I find the time, I will update mine to look more like this as well.

lakinwecker:
This looks good. Aside from fixing the one reply_method name and the 1/2Z-1/2Z - this is ready to go.  I will look into the test formatting stuff when you have merged this and we can deploy this along with the other stuff I have done today.

spreadsheets.js
endrawes0: I did not, so I may have missed a case -  rather just did this based on my experience. 

we will see if there are any oversights.

spreadsheets.js
endrawes0: I did not. Will remove.

spreadsheets.js
endrawes0: interesting. I am fine with either if you every decide to swap them out.

especially if this formatter handles things like precision

lakinwecker: If it's a replacement for the python one, it should handle precision, padding, etc.

chesster.js
```diff
@@ -1329,3 +1322,71 @@ controller.on('ambient', function(bot, message) {
...
+
+        }catch(e){
+            //at the moment, we do not throw from inside the api - rethrow
+            throw e;
+        }
+    });
+});
+
+function reply_permission_failure(bot, message){
+    bot.reply(message, "Sorry, you do not have permissin to update that pairing.");
```
lakinwecker: typo: permissin -> permission

PR #16, opened 2016-04-26
Fixing  few date parsing issues - also making it run faster

lakinwecker:
This version parses more date formats and runs slightly faster.  It accomplishes this by replacing weekday and month names and day-th (1st, 3rd etc) with the appropriate numbers and removing a bunch of other formats that we used to deal with specific formats. 

Also it removes punctuation at word-boundaries and refactors the methods to make them a bit more maintanable.

It fixes the tests to use a specific format and fixes the draw tests for 1/2Z and 1/2Z

### DIFF
--- a/spreadsheets.js
+++ b/spreadsheets.js
@@ -65,7 +65,8 @@ var BASE_TIME_FORMATS = [
     "HHmm",
     "Hmm",
     "HH-mm",
-    "H-mm"
+    "H-mm",
+    "HH"
 ];
 var DATE_FORMATS = [
 ];
@@ -208,6 +209,7 @@ function get_possible_date_strings(date_string, extrema) {
     month = now.format("MM");
     date_strings.slice().forEach(function(date_string) {
         date_strings.push("" + year + "-" + date_string);
+        date_strings.push(date_string + "-" + year);
     });
     return date_strings;
 }
--- a/spreadsheets.js
+++ b/spreadsheets.js
@@ -68,24 +68,22 @@ var BASE_DATE_FORMATS = [
     "D MMMM YY",
 ];
 var BASE_TIME_FORMATS = [
-    "HH:mm",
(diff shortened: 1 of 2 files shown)


spreadsheets.js
```diff
@@ -148,6 +119,101 @@ ScheduleParsingError.prototype = new Error();
...
+
+    var date_name_mappings = {};
+
+    var cur = extrema.start.clone();
+    var now = moment.utc();
+    while (!cur.isAfter(extrema.end)) {
+        var month_day = cur.format("MM-DD");
+        // Deal with a couple of formats that moment doesn't produce but are used.
+        if (cur.format("dddd") == "Thursday") {
+            date_name_mappings["thurs"] = month_day;
```
endrawes0: I don't understand why we have a special case for Thurs and Weds... You mean those are not supported abbreviations?

lakinwecker: That is correct. *dd* will produce: Mo, Tu, We, Th, Fr, Sa, Su.  ... *ddd* will produce the three-letter equivalents, but none of them produce Weds, or Thurs - which I've seen used.  

I agree, special cases are stupid, but not sure how else we should deal with this.

endrawes0: Understood. Thanks for the clarification. This version not only supports more formats but I think it's organized more clearly.

endrawes0:
I'm going to have lots of merging to do in my working copy.

Don't forget to close any issues that this solves.

lakinwecker:
Hopefully I didn't cause too much more work.

lakinwecker:
The only issue this addresses is the long running one that I have that tracks scheduling phrases that can't yet be parsed.  I'm leaving it open for now.

endrawes0:
We shall see how much. Won't be that bad.

Re. Issues: cool.

PR #19, opened 2016-04-28
Gets both config files current. Also removes any semblance of private keys

lakinwecker:
I realize this probably causes you to have more merges, but we should be diligent in keeping the config files up to date with new options and also ensuring that that the keys are not committed.  I'm considering moving the keys themselves to a completely different file that's _not_ in the repo.  

I haven't done that here though.  This is just getting stuff a bit more close to working out of the box to encourage collaborators.

### DIFF
--- a/chesster.json
+++ b/chesster.json
@@ -1,13 +1,14 @@
 {
-    "token":"",
+    "token":"YOU NEED ONE OF THESE - slack provides them, and you can use the modster one.",
+
     "service_account_auth": {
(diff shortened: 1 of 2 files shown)


endrawes0:
Yeah, my most recent version removed configuration from the repo but adds a sample. 

If I ever get around to finishing it...

lakinwecker:
Ok, then I will skip this pull request.

PR #23, opened 2016-04-30
Adds the pairing <player> <round> command for team

### LINKED ISSUE #15: make a my pairings command
It should start a DM and it should tell you both of your pairings, their rating, your scheduled times and/or results and perhaps a "countdown" till your game time:

example:

```
Lonewolf: you play endrawes0 (1687) on Sunday at 17:00 which is in 2day, and 3 hours.
Team: You played jaivl (1987) on Saturday at 08:00 and lost.
```

aweidner:
This attempts to close #15.  I was only able to add this for the team format but I think the modifications to make it work for lonewolf wouldn't be too difficult.  Also, I'm not sure right now how to format the times to be in the user's local time, so if I can get some guidance on that I would appreciate it!


endrawes0:
How was this tested? Also, who is this?!

endrawes0:
P.s. thanks for contributing

aweidner:
Haha yeah I should have put my username for the league on here so you didn't think I was some random dude.  I'm ghostologist, but didn't feel like making a new github account just for this stuff.  I set up a testing slack group and hooked the bot into that.  Then I tested with a few different rounds, ones that happend, ones that haven't been created yet, and the one that's currently in progress.  I also tested with captains and with regular players since they have different names.  I thought about possibly writing some unit tests for this stuff, but I didn't see an easy way to do it.

lakinwecker:
https://github.com/endrawes0/Chesster/blob/master/chesster.js#L1204

This shows you how to format the time for the given user. Of course, the time object should be created using moment.utc() so that it's a UTC time from the start.

chesster.js
```diff
@@ -644,6 +647,78 @@ controller.hears([
     });
 });
 
+controller.hears([
+    'pairing ([a-zA-Z0-9]+) ([1-8])'
```
lakinwecker: The first match is the player name right?  If so, you'll need to support more than just characters and letters. @r-mena had a dash in it, and it's not clear to me what other characters are allows. Also, if someone uses: pairing @lakinwecker, slack will translate that into @<USER-ID> which is different from the player name. 

If you use https://github.com/endrawes0/Chesster/blob/master/chesster.js#L97 this function, you can check to see if the word you are matching is a valid user - it will match both by name and id.  You'll have to strip off the @< ... > but otherwise it's pretty straight forward.

I'm starting to wonder if we should have a pre-set regex that we use to match a username or userID and a method that takes both, strips off the necessary characters and returns the appropriate user object.  Probably would be helpful.

chesster.js
```diff
@@ -644,6 +647,78 @@ controller.hears([
...
+                return;
+            } 
+        }
+
+        // If the round has not yet been scheduled, reply that and exit
+        callback("Round " + round + " has not been scheduled yet");
+    });
+}
+
+function parsePairingForRound(roundSheet, player, callback) {
```
lakinwecker: We have a "find_pairing"  method in spreadsheet.js - I'd like to see this use that, if possible. If that one is lacking - we should upgrade it so it supports both functionality.

In fact, it would be awesome if we were supporting finding pairings via only one side as that might help us locate and update schedules / results for people that played with alternates.

Any chance you can look at merging the two methods into one?

chesster.js
lakinwecker: You don't need to do all of the refactoring work that I suggest here. But I would like to see this also support the @lakinwecker syntax if possible.  People will use that often. And support dashes at the very least for people like @r-mena

chesster.js
```diff
@@ -644,6 +647,78 @@ controller.hears([
...
 
+controller.hears([
+    'pairing ([a-zA-Z0-9]+) ([1-8])'
+], [
+    'direct_mention',
+    'direct_message'
+], function(bot, message) {
+    bot_exception_handler(bot, message, function() {
+        var player_name = message.match[1];
+        var round = message.match[2]
```
lakinwecker: This should default to the latest round so that people can just issue: pairing @lakinwecker and get their current round pairing

chesster.js
endrawes0: We should, absolutely, predefine some filtering class that contains common utilities for parsing. Probably usernames, channels, etc.

We should add an issue for this refactoring.

chesster.js
endrawes0: Bump. Good suggestion.

endrawes0:
I've only had a chance to review @lakinwecker's suggestions. They are spot on as far as I can tell. 

I'll do a thorough review after good suggestions have been implemented or issues have been created.

aweidner:
I updated my pull request to support the latest syntax that was discussed earlier

I had a look at merging my function and find_pairing, but since it's involved in scheduling I don't want to muck around with it.  I don't have a direct route to validate that I'm not breaking anything if I touch that code right now

lakinwecker:
I can't seem to comment on specific lines, so I will comment here.  I think this is ready for @endrawes0 to review.  I have a couple small comments and one larger comment:

1. ```player = users.getByNameOrID(message.user);``` is repeated twice in the same method. A little bit wasteful, but probably the least of our worries in terms of our performance characteristics. :)
2. I wonder if this would be better as  DM response?
3. This doesn't work for lonewolf yet, and I think I'd like to see it work for both before we deploy it? Not, sure. I think we don't even have to detect which tournament the user is asking for, Just tell them about both. Especially if we are DMing them.  What do you think @endrawes0? I am leaning towards merging and deploying as is and then adding lone wolf as a second step

chesster.js
```diff
@@ -644,6 +647,95 @@ controller.hears([
...
+    'direct_mention',
+    'direct_message'
+], function(bot, message) {
+    bot_exception_handler(bot, message, function() {
+        // The user is either a string or an id
+        var nameOrId = message.match[1];
+        var requesting_player = users.getByNameOrID(message.user);
+
+        // The name or Id was provided, so parse it out
+        var player = null;
```
endrawes0: Is this line necessary? Why not not assign just once?

chesster.js
```diff
@@ -626,7 +629,7 @@ controller.hears([
 function preparePairingsMessage(){
     return "Here is the pairings sheet:\n" + 
             config.links.team + 
-            "\nAlternatively, try [ @chesster pairing <competitor> <round> ] - coming soon...";
+            "\nAlternatively, try [ @chesster pairing <competitor> <round> ]";
```
endrawes0: This is no longer correct. Round argument has been removed.

chesster.js
```diff
@@ -644,6 +647,95 @@ controller.hears([
...
+            bot.reply(message, response);
+        });
+    });
+});
+
+function uncaptain(playerName) {
+    return playerName.endsWith("*") ? playerName.substring(0, playerName.length - 1) : playerName;
+}
+
+function preparePairingCompetitorMessage(requesting_player, target_player, callback) {
```
endrawes0: We have not been consistent in using camel or snake case. This is not your problem. It's something that I need to address. Just noting it here.

endrawes0:
Overall this looks good. 

Please address my comments/questions and do a final pull from master - I just merged some changes from yesterday. 

I'll pull this as soon as the pull request has been updated.

Thanks! Good work!

chesster.js
endrawes0: Also, it should indicate that the <player> argument is optional.

chesster.js
```diff
@@ -655,26 +655,28 @@ controller.hears([
...
+        bot.startPrivateConversation(message, function (response, convo) {
+            // The user is either a string or an id
+            var nameOrId = message.match[1];
+            var requesting_player = users.getByNameOrID(message.user);
 
-        // The name or Id was provided, so parse it out
-        var player = null;
-        var player = users.getByNameOrID(nameOrId);
+            // The name or Id was provided, so parse it out
+            var player = null;
```
endrawes0: can you comment on why we need this line?

chesster.js
```diff
@@ -645,6 +648,97 @@ controller.hears([
...
+    'direct_message'
+], function(bot, message) {
+    bot_exception_handler(bot, message, function() {
+        bot.startPrivateConversation(message, function (response, convo) {
+            // The user is either a string or an id
+            var nameOrId = message.match[1];
+            var requesting_player = users.getByNameOrID(message.user);
+
+            // The name or Id was provided, so parse it out
+            var player = null;
```
endrawes0: can you comment on why this line is necessary?

aweidner: it is not necessary, I think it's left over cruft from my development.  I should really pay closer attention to that.  Thanks for pointing it out.  I will remove it and resubmit!

PR #24, opened 2016-05-01
Gamelink parsing

endrawes0:
This is  relatively large pull request. For the most part it introduces gamelink parsing but it also fixes some punctuation issues in results parsing.

The implementation can be made more efficient. At the moment, I chose simplicity of implementation and functional decomposition over efficiency.

When we start caching the row result we should be fast enough that the efficiency issues are not noticeable.


chesster.js
```diff
@@ -1351,7 +1349,10 @@ controller.on('ambient', function(bot, message) {
             return;
         }
         try{
-            var result = spreadsheets.parse_result(message.text, results_options);
+            var result = 
+                spreadsheets.parse_result(
+                    message.text, 
+                    config.gamelinks[channel.name]);
```
lakinwecker: Why not use ```results_options``` here?

endrawes0: You are correct. I should be. This would have led us to parsing the results in the wrong channel. Thanks for pointing it out.

chesster.js
```diff
@@ -1365,23 +1366,48 @@ controller.on('ambient', function(bot, message) {
...
-                        result_reply_updated(bot, message, result);
+                result,
+                function(err, gamelink){
+                    //if a gamelink is found, use it to acquire details and process them
+                    if(!err && gamelink){
+                        process_gamelink(
+                            bot, 
+                            message, 
+                            gamelink, 
+                            config.gamelinks[channel.name]);
```
lakinwecker: What about here too?

endrawes0: Not here. Here we need the game link options. 

Actually, I need to double check both of those assertions. I know we need separate options because the channels are different but the correct case to use each is not as clear to me now as it was last night.

I'll verify and let you know.

endrawes0:
I will use ⚔ as an active game marker in the spreadsheet. This still needs to be done. Will update later

spreadsheets.js
```diff
@@ -506,16 +506,19 @@ function find_players(tokens){
...
     }
 
     return players;
 }
 
 function filter_player_tokens(tokens){
     return _.filter(tokens, function(token){
-        return /^\<@[A-Z0-9]+\>$/.test(token);
+        //matches slack uer ids: <@[A-Z0-9]>[:]*
+        return /^\<@[A-Z0-9]+\>[:]*$/.test(token);
```
lakinwecker: This needs to support a dash as well. @r-mena as the example.  I'll suggest we use the same one that I suggested for the other PR: [^\s]+ which should match all non-white-space characters.

spreadsheets.js
```diff
@@ -537,37 +541,109 @@ function update_result(service_account_auth, key, colname, result, callback){
...
+            
+            var gamelink;
+            if(!gamelink_id && formula && formula.toUpperCase().includes("HYPERLINK")){
+                //there is no gamelink, fill from the formula
+                gamelink = formula.split("\"")[1];
+            }else if(!result.gamelink_id){
+                //no gamelink and no formula
+                gamelink = '';
+            }else{
+                gamelink =  "http://en.lichess.org/" + result.gamelink_id;
```
lakinwecker: Constructing this URL is done in two places. not a big deal, but could get unified.

spreadsheets.js
```diff
@@ -537,37 +541,109 @@ function update_result(service_account_auth, key, colname, result, callback){
...
                 return callback(err, reversed);
             });
         }
     );
 }
 
+//given the input text from a essage
+function parse_gamelink(message_text){
+    //split it into tokens separated by white space and slashes
+    var tokens = message_text.split(/[\/ \t\n\r]/);
```
lakinwecker: ```\s``` represents a single whitespace character.

chesster.js
```diff
@@ -1401,3 +1427,179 @@ function result_reply_missing_pairing(bot, message){
...
+    bot.reply(message, "Something went wrong. Notify " + contact);
+}
+
+function gamelink_reply_unknown(bot, message){
+    bot.reply(message, "Sorry, I could not find that game. Please verify your gamelink.");
+}
+
+function process_gamelink(bot, message, gamelink, options){
+    //get the gamelink id if one is in the message
+    var result = spreadsheets.parse_gamelink(gamelink, options);
```
lakinwecker: parse_gamelink does not use the options as of yet. Yet you are passing it here. Is that an oversight, or are you planning to make it use the options in the future?

chesster.js
lakinwecker: You are correct. It should be config.gamelinks options here. This is eventually passed into the function that validates the game results, so it needs this version.

chesster.js
lakinwecker: In this case, parse_result doesn't actually use the options either. So we could just remove it.

chesster.js
```diff
@@ -1401,3 +1427,179 @@ function result_reply_missing_pairing(bot, message){
...
+
+
+
+/* game link parsing */
+
+//given a gamelink_id, use the lichess api to get the game details
+//pass the details to the callback as a JSON object
+function fetch_game_details(gamelink_id, callback){
+    const http = require('http');
+    var url = "http://en.lichess.org/api/game/" + gamelink_id;
```
lakinwecker: I feel like we could make use of a function that just reads a result into a JSON Structure in memory. And this is the only line here that's not about that aspect.  I'd be in favour (but it's not required for this PR) to refactor this into a "fetch_url_into_json" method and to then pass in a complete url. 

you could keep a fetch_game_details method around that takes only a gamelink_id, but it should use the above method internally.

chesster.js
```diff
@@ -1365,23 +1366,48 @@ controller.on('ambient', function(bot, message) {
...
-                            bot.reply(message, "Something went wrong. Notify @endrawes0");
-                            throw new Error("Error updating scheduling sheet: " + err);
-                        }
-                    } else {
-                        result_reply_updated(bot, message, result);
+                result,
+                function(err, gamelink){
+                    //if a gamelink is found, use it to acquire details and process them
+                    if(!err && gamelink){
+                        process_gamelink(
```
lakinwecker: In this case specifically, we might want to actually validate the result against what is reported by the gamelink. I know there was at least one game last season that the lichess result was incorrect, as the players had verbally agreed to a draw, but the resign button was pressed. In the case where the results differ from the gamelink, we should inform them, and not update the spreadsheet.

I am ok with this PR being accepted and deployed without this feature - but I would like to see this feature added in soon afterwards.

lakinwecker:
This is good work. A few small things I would like to see cleaned up before we push it live, like the updated regex for usernames that supports r-mena, as well as the removal of the options parsing where it's not necessary.  The rest are all suggestions that are not required for this PR to be merged and deployed.

This is just large enough and just difficult enough to verify by reading code that I would like to be around for the deployment to do some final testing of a few edge cases before we turn it lose on the channels. 

Good job though!  I'm super excited to get this deployed as well.

chesster.js
endrawes0: Removed this parameter.

spreadsheets.js
endrawes0: This is expecting the slack user id and markup.

It should not contain the user's handle. Are you suggesting I use this to removing trailing punctuation?

spreadsheets.js
endrawes0: Where else is it done? I cannot find any other references.

chesster.js
endrawes0: Oversight - same as the parse_result above. Not sure why. Will remove/

spreadsheets.js
lakinwecker: You are correct. Carry on.

chesster.js
lakinwecker: This is where I was thinking it was done again, but this is a different call. Again, you are correct. Carry on.

endrawes0: completely agree with this refactoring. done.

chesster.js
endrawes0: OK - I have implemented this. I agree it is important. I will need to do more testing to make sure it is working correctly.

spreadsheets.js
endrawes0: I am  using \s to identify extra characters at the end of the token

e.g.

@endrawes0.
@endrawes0,
@endrawes0:
@endrawes-

These should all be identified.

endrawes0:
I have made the changes you have suggested. I still need to do some testing before I recommit.

spreadsheets.js
endrawes0: oh - ended up removing the \s to identify non whitespace at the end. I could not get the regex right. Maybe we can review another time. 

I looked at something like

/^\<@[A-Z0-9]+\>!([\s])*$/

soemthing wrong with the syntax I think

chesster.js
```diff
@@ -1507,15 +1509,55 @@ function gamelink_reply_unknown(bot, message){
...
+        validity.reason = "the user claimed a draw " 
+                        + "but the gamelink specifies " + details.winner + " as the winner.";
+        validity.valid = false;
+   }else if( details.winner == "black" && result.result == "1-0"){
+        //the details gave the winner as black but the user claimed white
+        validity.reason = "the user claimed a win for white " 
+                        + "but the gamelink specifies black as the winner.";
+        validity.valid = false;
+    }else if( details.winner == "white" && result.result == "0-1"){
+        //the detauks gave the winner as white but the user claimed black
```
lakinwecker: typo. detauks -> details.

spreadsheets.js
lakinwecker: Yeah, let's make a ticket for getting these regexes unified so we don't have to keep writing them for each individual spot.

lakinwecker:
This looks good to go.

PR #28, opened 2016-05-02
replace date with moment and fix gamelink parsing when it comes in as markup

endrawes0:
no hurry on this.

its working as is, but this is for consistency


endrawes0:
added a commit to handle gamelinks as markup

lakinwecker:
This all looks good for me.

Issue #32, opened 2016-05-05
Typo - Because is spelled incorrectly


lakinwecker:
https://github.com/endrawes0/Chesster/blob/master/chesster.js#L1607

endrawes0:
Ack. Thanks for noticing

endrawes0:
#34 fixed this.

PR #34, opened 2016-05-07
Fixes because typo and rating @<player> syntax

### LINKED ISSUE #21: @chesster rating @<player> fails

aweidner:
This fixes #21 and #32.  I factored out the code to pull a player id so we can generate commands that take players as an argument and it should work with:

player
@player
playerMixedCase
aj;sldfkajs;dlfkj (garbage)
<blank>

### DIFF
--- a/chesster.js
+++ b/chesster.js
@@ -7,6 +7,7 @@ var fuzzy = require('./fuzzy_match.js');
 var spreadsheets = require('./spreadsheets.js');
 var http = require('http');
 var moment = require('moment');
+var players = require("./player.js");
 
 var MILISECOND = 1;
 var SECONDS = 1000 * MILISECOND;
@@ -300,14 +301,13 @@ function sayCaptains(self, convo, callback){
 /* rating */
 
 controller.hears([
-	'rating'
+    players.appendPlayerRegex("rating", true)	
 ],[
 	'direct_mention', 
 	'direct_message'
 ],function(bot,message) {
     bot_exception_handler(bot, message, function(){
-        var self = this;
-        var player_name = message.text.split(" ").slice(1).join(" ");
+        var player_name = players.getSlackUser(users, message).name;
         if(player_name && player_name != ""){
             getRating(player_name, function(rating){
                 if(rating){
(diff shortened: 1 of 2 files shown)


chesster.js
```diff
@@ -7,6 +7,7 @@ var fuzzy = require('./fuzzy_match.js');
 var spreadsheets = require('./spreadsheets.js');
 var http = require('http');
 var moment = require('moment');
+var players = require("./player.js");
```
endrawes0: Yes. More of this please 👍 

We should continue to make minor improvements like this.

chesster.js
```diff
@@ -300,14 +301,13 @@ function sayCaptains(self, convo, callback){
 /* rating */
 
 controller.hears([
-	'rating'
+    players.appendPlayerRegex("rating", true)
```
endrawes0: with this change I think we can remove the error handling below.

it will no longer match if there is no player

endrawes0: I cant put a comment on the line so I'll note it here: error on line 320 can be removed.

chesster.js
```diff
@@ -1604,7 +1581,7 @@ function validate_game_details(details, options){
 
 function gamelink_reply_invalid(bot, message, reason){
     bot.reply(message, "I am sorry, <@" + message.user + ">,  "
-                     + "the link you posted is *not valid* becuase "
+                     + "the link you posted is *not valid* because "
```
endrawes0: thank you.

player.js
```diff
@@ -0,0 +1,37 @@
+module.exports.appendPlayerRegex = function(command, optional) {
+    var playerRegex = command + "(?: @?([^\\s]+))";
```
endrawes0: can you put a comment in that explains the regex? thank you :)

player.js
```diff
@@ -0,0 +1,37 @@
...
+
+    if (optional) {
+        return new RegExp(playerRegex + "?");
+    }
+
+    return new RegExp(playerRegex);
+};
+
+module.exports.getSlackUser = function(users, message) {
```
endrawes0: doesnt need to happen now, but probably lichess user info could be added here... 

OR maybe we move all the lichess stuff into its own js file. in any case, more refactoring and better code organization.

endrawes0:
Hey Ghost. 

I made a few comments/requests.

Please take a look. 

Thanks.
Andrew

PR #38, opened 2016-05-14
redirect to logfiles in start script.

endrawes0:


start
```diff
@@ -1,2 +1,2 @@
 #!/bin/bash
-node chesster.js $1
+node chesster.js $1 >> stdout_log 2>> stderr_log
```
aweidner: I believe these files should be added to the .gitignore.

aweidner:
One thing to consider is that chesster shouldn't currently write anything to stderr.  I couldn't find any uses of `console.error` in the codebase, everything is currently printing to stdout through `console.log`.     There's also a file called `errors_log` which looks like it contains stacktraces.  What should go into the stderr_log file and what should go into errors_log, do we need both?

start
lakinwecker: I think I'd prefer to have a single log for all of it, something like

```2&>1 >> output.log```

I don't want to have to remember which log to look at for which information.

lakinwecker:
Long term we may want to switch to a proper logging library, but I think a simpler fix, like this, should be good for now

Example Logging library:  https://github.com/winstonjs/winston

start
endrawes0: Hm, I'm ok with that - any logging system your are familiar with? This goes above and beyond the requirements, but could be good to tag the beginning of each line with a timestamp and file?

`2016-05-14T16:13:58+00:00 - stdout: something something something `

endrawes0:
Just saw the reference to Winston

start
endrawes0: Agreed about the ignore

endrawes0:
In my future commits, I will utilize console.error and replace the error log.

for now, just redirecting

endrawes0:
updated and ready for merging if there are no other issues.

lakinwecker:
:+1:

endrawes0:
ok - I lied. I needed to add to .gitignore.

PR #40, opened 2016-05-14
Refactoring and tests

lakinwecker:
NOTE: this is not a pull request that's ready to be merged. I am making it so that we can easily see and discuss the way in which I am making these changes to see if people like the ideas.  

The idea is that we would make a league object which would abstract away dealing with the differences between the spreadsheets, so that commands could operate on both leagues or one league and not have to worry about which one they are using.

I would like to also add a way to register a controller.hears type message which will automatically pass in the league instance based on which channel you are using.  

This also starts down the path of caching information from the spreadsheets to make the actual commands faster and simpler.


config.js
```diff
@@ -0,0 +1,75 @@
+// NOTE: Neither of these files are committed and for good reason.
```
endrawes0: will configuration json (ie chesster.json and tesster.json) still be something loaded from the file as well, or just a config.js.

seems we shouldnt hard code this right?

lakinwecker: I'm planning to make which config.js you load come from the command line, just like it used to. If that's the case, you can have a tesster.js which loads the appropriate versions of these. 

Alternatively, it doesn't matter. I never want the live token/keys on my dev environment, and I don't want the testing ones in the same directory structure on the live environment.  So I'm not sure how important it is?  

Even having said that, you're probably right. I should probably make it easier to switch between the two.

league.js
```diff
@@ -0,0 +1,167 @@
...
+    }
+};
+
+function League(options) {
+    this.options = {};
+    _.extend(this.options, LEAGUE_DEFAULTS, options || {});
+    _.extend(this, league_attributes);
+};
+
+var _cache = {};
```
endrawes0: I suggest moving cache into the getLeague function to avoid scope pollution
```
function getLeague(league_name, config) {
    this._cache = ( this.cache ? this.cache : {} );

    if(!this._cache){
        console.log(league_name);

        // Else create it if there is a config for it.
        var league_options = config[league_name] || undefined;
        if (league_options) {
            this._cache[league_name] = new League(league_options);
        }else{
            console.log("Couldn't find options for " + league_name + " league. Not creating object.");
        }
    }
    return this._cache[league_name];
}
```

Also a little refactoring.

lakinwecker: Your point stands, but your version will only work for the first league that's created. Subsequent calls will get undefined for new league names.

chesster.js
```diff
@@ -161,6 +152,17 @@ function update_channels(bot){
     });
 
 }
+function refreshCurrentRoundSchedules() {
+    var _45_45 = league.getLeague("45+45", config);
+    _45_45.getCurrentRoundSchedule(function(err, pairings) {
```
endrawes0: I like where this is going, but since the leagues will likely update more than round schedules, I think we should simply have a `refresh` function

```
var _45_45 = leage.getLeague("45+45", config);
_45_45.refresh();
```

where `refresh` calls `getCurrentRoundSchedule`, `getMods`, `getRosters`, etc

lakinwecker: I agree. Although I think that the league will want to control how often each gets called.  For instance, getRosters only needs to be called once per week (ish) while getCurrentRoundSchedule will need to be called far more often, like it is here.

Those times could come from the config and perhaps it's more of a:

```_45_45.installRefreshTimeouts()``` or something?

File: league.js [hunk 4]
```diff
@@ -0,0 +1,167 @@
...
+        return username.replace("*", "");
+    },
+
+    //--------------------------------------------------------------------------
+    // Syncs the latest team information
+    //--------------------------------------------------------------------------
+    'get_rosters': function(callback) {
+    },
+
+    //--------------------------------------------------------------------------
```
endrawes0: we should design on how the functions will be named

`getMods` should update the cache
`findMods` should check the cache first

in general, a get will hit the spreadsheets and a find will try the cache first and then hit the spreadsheet if it fails
this way all the functions can be implemented in terms of one another

I dont really care what the exact names are as long as we have a standard and we follow it.

lakinwecker: Agreed.

chesster.js
lakinwecker: Actually, I have changed my mind. I like your version.

While getRosters doesn't _need_ to run every 30s, we don't know when the mods will actually update the rosters, so we can't time the refresh to happen immediately after that. So it's probably fine to just update all three at the same interval

config.js
endrawes0: I agree entirely regarding segregation of the keys.

The concern was whether or not the config could be replaced easily.

I dont want to have to push changes to chesster when the league format changes - unless it changes in a way that requires changing the format of the config - that is a different story - but this format of config seems flexible enough.

league.js
endrawes0: oh - not even for the first one, would a value be found - the code was meant to read:

```
function getLeague(league_name, config) {
    this._cache = ( this.cache ? this.cache : {} );

    if(!this._cache[league_name]){
        console.log(league_name);

        // Else create it if there is a config for it.
        var league_options = config[league_name] || undefined;
        if (league_options) {
            this._cache[league_name] = new League(league_options);
        }else{
            console.log("Couldn't find options for " + league_name + " league. Not creating object.");
        }
    }
    return this._cache[league_name];
}
```

chesster.js
endrawes0: you make a good point - we may want to make refresh of particular entities possible from slack.

this way when theino or mrlegs are updating pairings, they can send `@chesster refresh pairings`

endrawes0:
As I said in an inline comment, I like where this is going. Some more discussion - and I would like to hear what @aweidner thinks - then I am on board.

chesster.js
lakinwecker: That is a good idea.

config.js
lakinwecker: I see your point.

league.js
lakinwecker: In this case, I'm going with "refreshX" for refreshing it from the spreadsheets.  And "findX" for when you're doing a non-exact search and "getX" for when you're asking for _exactly_ this item.

Neither findX or getX necessarily talk to the spreadsheet.  Does that work?

lakinwecker:
Thanks for the review. I'm going to make some of the changes you suggested, but otherwise wait for @aweidner

league.js
```diff
@@ -0,0 +1,201 @@
...
+                callback(undefined, self._pairings);
+            }
+        );
+    },
+
+    //--------------------------------------------------------------------------
+    // Finds the pairing for this current round given either a black or a white
+    // username.
+    //--------------------------------------------------------------------------
+    'findPairing': function(white, black) {
```
aweidner: This can be refactored no?  Something like:

    filterByColor(white, possibilities)
    filterByColor(black, possibilities)
 
    ...

    function filterByColor(color, possibilities) {
        if (color) {
              possibilities = _.filter(possibilities, function(item) {
                  return (
                      item.white.toLowerCase().includes(color) ||
                      item.black.toLowerCase().includes(color)
                  );
              });
          }
    }

If you get my meaning

aweidner:
Overall, yes this looks good to me.  So the general idea is that we just have to create configs for the two different leagues, and then commands that have been ported to work from the league object will work with both out of the box correct?

Caching the spreadsheet information is going to be really good.  I know for instance that the pairing command is really slow because we have to hit google and lichess apis.  If we just have to request from lichess that's going to make things a lot snappier.  Any plans to cache the lichess information?  Is this feasible?

league.js
lakinwecker: Yup, that makes sense to me.

lakinwecker:
Yeah, I think we should probably cache certain things from lichess like rating.  We can't cache game information, but it still should be better this way.

The generally idea is correct as well. Ideally this league object would know how to work appropriately with the leagues and the bot handlers wouldn't have to worry about those details.

league.js
```diff
@@ -0,0 +1,296 @@
...
+
+function getAllLeagues(config) {
+    var leagues = [];
+    var all_league_configs = config['leagues'] || {};
+    _.each(_.keys(all_league_configs), function(key) {
+        leagues.push(getLeague(key, config));
+    });
+    return leagues;
+}
+
```
lakinwecker: @endrawes0 Note, I tried your version but the this was different for whatever context you called it in. At least that was my assessment when this function was creating the same league twice from two different contexts.  This has a similar effect (the _league_cache is not in the global scope), but is more explicit about the scope and ensures that the _league_cache is effectively static (one copy per process).

chesster.js
```diff
@@ -636,115 +547,39 @@ controller.hears([
     });
 });
 
-controller.hears([
+slack.hears(controller, [
```
endrawes0: I definitely like the simplifications that slackjs provides on the `hear` functions. Alot of redundancy will be removed. Fantastic.

chesster.js
```diff
@@ -1,13 +1,22 @@
+// extlibs
 var async = require("async");
 var Botkit = require('botkit');
```
endrawes0: Botkit is now required twice - once here and once by slackjs. Any impact? If so, any way to refactor this so it's only required by slackjs?

I'm thinking it should be feasible but would mean moving the controller initialization into slackjs. It would also simplify our hears functions because controller would not need to be passed in.

Instead we would have a slack object which maintains the state that controller normally would.

lakinwecker: There shouldn't be any impact as far as I know.  

If we get to a point where it's possible to remove the botkit reference in chesster.js, then by all means let's do it. However, that won't be my goal within this pull request.  As i don't want this branch to live for too long without being merged.

chesster.js
lakinwecker: Yeah, I'm pretty happy with that aspect as well.

chesster.js
```diff
@@ -636,115 +522,63 @@ controller.hears([
...
+    message_types: [
+        'direct_mention', 'direct_message'
+    ],
+    config: config,
+}, function(bot, message) {
+    var targetPlayer = players.getSlackUser(users, message);
+    var deferred = Q.defer();
+    var allLeagues = league.getAllLeagues(config);
+    bot.startPrivateConversation(message, function (response, convo) {
+        Q.all(
```
endrawes0: As I understand it, `Q.all(...)` is meant to return an array of results. Are we using the array result somewhere?

Do we need an array result that I dont see?

If not, is this the right method for our use case?

endrawes0: Seems we are just using the promise for control flow. Yeah?

lakinwecker: Q.all returns a promise which will wait until all of the given promises have completed before it completes. Similar to async.parallel except for promises.

The promise is being used for flow control here, yes. but that's not the only reason it's being used. promises are now used throughout this paradigm. Happy to chat more about my use of them if desired.

league.js
```diff
@@ -0,0 +1,323 @@
...
+            }
+        );
+    },
+
+    //--------------------------------------------------------------------------
+    // Finds the pairing for this current round given either a black or a white
+    // username.
+    //--------------------------------------------------------------------------
+    'findPairing': function(white, black) {
+        white = white || undefined;
```
endrawes0: is this an idiom I am not familiar with?

Essentially, if white has a value that will evaluate to true, use it, otherwise assign undefined.

It will replace 0, false, null, etc with undefined.

Anything I missed? Why make this assignment?

league.js
```diff
@@ -0,0 +1,323 @@
...
+    // Generates the appropriate data format for pairing result for this league.
+    //--------------------------------------------------------------------------
+    'getPairingDetails': function(targetPlayer) {
+        var self = this;
+        return Q.fcall(function() {
+            var pairings = self.findPairing(targetPlayer.name);
+            if (pairings.length < 1) {
+                return {};
+            }
+            // TODO: determine what to do if multiple pairings are returned. ?
```
endrawes0: If multiple pairings are returned... 

Its either a very special case or something is broken.

In the former case, we can probably just ignore it (say if we update the round with someone playing up a board and they get 2 matches in a single round). In the latter case we would want to notify a moderator (say in the case that data entry is bad).

Erring on the side of caution would mean alerting a moderator in this case.

league.js
lakinwecker: Your interpretation of what the assignment does is correct.  

Also, you are not missing anything. I added this (and the version for black) in a previous (uncommitted) incarnation of this function. I'm not even certain it provided value in that version, but it certainly does not provide any value here.  I will remove.

league.js
lakinwecker: Yeah, I agree with everything you said - but I'm still uncertain of what the right course of action is in that case.  I'm happy to implement any of the implied solutions: ignore it, include it, and/or notify a mod.

league.js
endrawes0: Yes - this is actually a nice variant of implementation but achieving static-ness simply by maintaining a references via the functions that access the cache.

Its downright neat. Good idea.

slack.js
```diff
@@ -0,0 +1,266 @@
+//------------------------------------------------------------------------------
+// Bot / Slack related helpers
+//------------------------------------------------------------------------------
+var Botkit = require('botkit');
```
endrawes0: I dont actually see a use of Botkit in slackjs.

The reference is passed in from the caller.

I believe this can be removed.

league.js
endrawes0: I suggest Including multiples and telling the caller that they should notify the moderators if this is unexpected.

chesster.js
endrawes0: Nothing to discuss at the moment. Just making sure I understood how/why it was used here. You have confirmed my understanding. Thank you.

slack.js
```diff
@@ -0,0 +1,266 @@
...
+    byId: {},
+    getId: function(name){
+        return this.byName[name].id;
+    },
+    getIdString: function(name){
+        return "<#"+this.getId(name)+">";
+    }
+};
+
+function updateUser(bot){
```
endrawes0: `updateUsers` I think

slack.js
endrawes0: alternatively, we could internalize botkit in slackjs. 

either way is fine with me.

slack.js
```diff
@@ -0,0 +1,266 @@
...
+        });
+        setTimeout(function(){ 
+            refresh(bot, delay, config);
+        }, delay);
+    }));
+}
+
+function criticalPath(promise){
+    exceptionLogger(promise).catch(function(e) {
+        console.log("An exception was caught in a critical code-path. I am going down.");
```
endrawes0: we should print `e` here.

endrawes0: and this should be `console.error`

I believe this was just copy pasted from the old implementation so feel free to not touch it. just pointing it out.

lakinwecker: exceptionLogger already logs e. I have changed console.log to console.error in my local copy as well.

slack.js
```diff
@@ -0,0 +1,266 @@
...
+            throw new Error("requiresModerator MUST be called after withLeague.");
+        }
+        if (!message.league) {
+            throw new Error("Not in a league context");
+        }
+        var item = _.find(message.league.options.moderators, function(moderator) {
+            return message.player.name == moderator;
+        });
+        if (typeof item == 'undefined') {
+            bot.reply(message, "You are not a moderator of the {name} league. Your temerity has been logged.".format({
```
endrawes0: oh boy. caught in the act!

slack.js
endrawes0: the more and more I read this code, the more I think internalizing the botkit api would be a good idea.

it will simplify the interface if the client doesnt need to know about botkit and we can wrap up all the botkit specific code within one file.

What are your thoughts?

lakinwecker: Long term, I agree. But for this pull request I'd prefer to hold off on that work. 

I don't have the time to rewrite every command in the league to use this new paradigm, so I'm hoping I can just fix the ones that are broken by the new config format and then we can merge with master.  After that we can more slowly port over the rest of the commands and once they are ported, then we can fully internalize it.  

And until we fully internalize it, some of the commands in chesster.js and some in slack.js will need access to things, so we'll end up with a less ideal interface for using slack.js.  But we can fix that when we finish the porting work. 

If you're keen on getting it all ported over and are available to help, then I'm happy to start moving it that way.

chesster.js
```diff
@@ -62,268 +47,89 @@ function critical_path(todo){
...
-],[
-	'ambient'
-],function(bot,message) {
-    bot_exception_handler(bot, message, function(){
-        var response = "Stop... Giri Time!\n" + "Hi! Im Chesster. Ill be your new bot. " + 
-						"To interact with me, mention " + users.getIdString("chesster") + 
-						" in a message";
-        bot.reply(message, response);
-    });
+chesster.hears({
```
endrawes0: ha. I appreciate you leaving this here. I can definitely clean this up. It was essentially my "Hello, World!"

chesster.js
```diff
@@ -62,268 +47,89 @@ function critical_path(todo){
...
+},
+function(bot,message) {
+    var response = "Stop... Giri Time!\n" + "Hi! Im Chesster. Ill be your new bot. " + 
+                    "To interact with me, mention " + slack.users.getIdString("chesster") + 
+                    " in a message";
+    bot.reply(message, response);
 });
 
 /* captains */
+leagueResponse(['captain guidelines'], 'formatCaptainGuidelinesResponse');
```
endrawes0: This requires a league? What does it do if raised in #general channel?

Ill probably find the answer to this later but just in case i dont, I am asking now.

lakinwecker: Yes, it requires a league. If we ever get a second league that has captains, then this would need to know which league to respond with.  Also, the config for this is league-specific now.  

We can make the #general channel be a a team league channel and then this would work in that channel. Or we could improve the requiresLeague to respond to someone that issues a league-command with an request for them to identify the league they are interested in.

chesster.js
```diff
@@ -62,268 +47,89 @@ function critical_path(todo){
...
+},
+function(bot,message) {
+    var playerName = players.getSlackUser(users, message).name;
+    return lichess.getPlayerRating(playerName).then(function(rating) {
+        if(rating){
+            bot.reply(message, prepareRatingMessage(playerName, rating));
         }else{
-            console.error(JSON.stringify(error));
-            callback();
+            bot.reply(message, "I am sorry. I could not find that player.");
```
endrawes0: instead of implementing this as an else clause of the conditional, could we use the second argument of the `then(...)` function to send the response? 

Not sure if maybe the second argument is reserved for other kinds of errors or if this is a valid use-case.

chesster.js
```diff
@@ -365,190 +171,34 @@ function prepareCommandsMessage(){
...
+        'command list',
+        'help',
+    ],
+    messageTypes: [
+        'direct_mention', 
+        'direct_message'
+    ]
+},
+function(bot,message) {
+    bot.startPrivateConversation(message, function (response, convo) {
```
endrawes0: this should use the exception handler, no? Are we still using it everywhere? Is it still necessary?

chesster.js
```diff
@@ -365,190 +171,34 @@ function prepareCommandsMessage(){
...
-            }
-        } else {
-            console.log("Unable to determine target");
-        }
-    });
-});
+leagueResponse(['summon mods'], 'formatSummonModsResponse');
+leagueResponse(['mods'], 'formatModsResponse');
 
 /* help */
```
endrawes0: this comment is not needed any more

chesster.js
lakinwecker: It's not reserved. I guess it's a matter of deciding if it's an error to ask for a rating of a player that does not exist or not.    I'm undecided.

chesster.js
```diff
@@ -617,446 +259,128 @@ controller.hears([
...
-controller.hears([
-    'standings'
-],[
-    'direct_mention', 
-    'direct_message'
-],function(bot, message) {
-    bot_exception_handler(bot, message, function(){
-        bot.reply(message, prepareStandingsMessage());
+chesster.hears({
+    middleware: [slack.requiresLeague, slack.requiresModerator],
```
endrawes0: :heart: :heart: :heart: I love this... :heart: :heart: :heart:

chesster.js
lakinwecker: chesster.hears does wrap this in a bot exception handler.  However, just like all callback code, that exception handler does _not_ deal with exceptions that occur inside of the next level of callbacks.  This was the case with the previous version as well as this version, so it's no worse than before.
 
However, the call to ```bot.startPrivateConversation``` could be wrapped in a Q.nfcall and it would be properly protected via the promise chain.  We should do that.

chesster.js
lakinwecker: Yeah, being able to quickly write commands that require a moderator is pretty awesome.

chesster.js
```diff
@@ -617,446 +256,128 @@ controller.hears([
...
-                            }
-                        });
-                    }else{
-                        bot.reply(message, "I could not find that team.");
-                    }
-                });
+/* team members */
+chesster.hears({
+    middleware: [slack.requiresLeague],
+	patterns: ['team members'],
```
endrawes0: indentation

chesster.js
```diff
@@ -617,446 +256,128 @@ controller.hears([
...
-controller.on('user_channel_join', function(bot, message) {
+chesster.on({event: 'user_channel_join'},
+function(bot, message) {
     bot_exception_handler(bot, message, function(){
         if(message.channel == channels.getId("general")){
             bot.reply(message, "Everyone, please welcome the newest member of the " 
                              + "Lichess 45+45 League, <@" + message.user + ">!");
             
             bot.startPrivateConversation(message, function(err, convo){
+                // TODO: the config links references below are hard coded
```
endrawes0: Were you thinking all leagues could be displayed or perhaps a more interactive menu that would dive in to a particular league's links?

chesster.js
```diff
@@ -1071,8 +392,8 @@ controller.on('user_channel_join', function(bot, message) {
                         + "connected with the league. It is the easiest way for most "
                         + "of us to communicate and you will find that many of us "
                         + "are active in this community every day. Make yourself at home.");
-                convo.say(config.links.guide);
-                convo.say(config.links.rules);
+                convo.say(chesster.config["leagues"]["45+45"].links.guide);
```
endrawes0: easy to throw this in a map all and hit all the leagues if it isnt too obnoxious with many leagues

chesster.js
lakinwecker: When I wrote this, I wasn't sure. I was, however, trying to ensure that no league-specific stuff was hard coded in chesster.js and this is 45+45 league specific.  

Since then I've thought about this and realized that this probably needs to change. Lone Wolf signups are being confused by this message when they join when the team tournament is not active.  I would be in favour of making this much shorter and much more about the very basics:

You're new here, We're happy you joined us. We have two leagues and I'm not sure which one you signed up for, but I'd like to tell you about the league you joined. If you want an introduction with the 45+45 team tournament, say: intro team if you want an introduction to the lone wolf tournament, say: intro lonewolf.

And then there would be an 'intro' command that was league specific. And the number of leagues that have intros could be determined at runtime.

chesster.js
lakinwecker: That is also an approach I would be happy with.

chesster.js
```diff
@@ -1312,224 +516,244 @@ function scheduling_reply_scheduled(bot, message, results, white, black) {
...
+        isPotentialSchedule = true;
+    } catch (e) {
+        if (e instanceof (spreadsheets.ScheduleParsingError)) {
+        } else {
+            throw e; // let others bubble up
         }
+    }
 
+    // Unless they included a date we can parse, ignore this message.
+    if (!isPotentialSchedule) {
```
endrawes0: Why dont we just return in the catch? seems this will only false if the catch is executed.

chesster.js
```diff
@@ -1312,224 +516,244 @@ function scheduling_reply_scheduled(bot, message, results, white, black) {
...
+                if (err.indexOf && err.indexOf("Unable to find pairing.") == 0) {
+                    hasPairing = false;
                 } else {
-                    has_pairing = true;
-                }
-                if (reversed) {
-                    var tmp = white;
-                    white = black;
-                    black = tmp;
+                    bot.reply(message, "Something went wrong. Notify @lakinwecker");
```
endrawes0: could probably just say "Notify a mod." if you want to give yourself a break.

chesster.js
lakinwecker: Probably. But this can go into a ticket?  I'm not sure I want to change this algorithm in the middle of this refactor.

config.js
```diff
@@ -0,0 +1,175 @@
...
+//       You must provide your own.
+var token = require("./slack_token.js").token;
+var private_key = require("./service_account_key.js").key;
+
+var config = {
```
endrawes0: not sure if I mentioned this before, but I like the way you have reorganized this.

config.js
```diff
@@ -0,0 +1,175 @@
...
+            "links": {
+                "rules": "https://lichess4545.slack.com/files/parrotz/F0D7RD88L/lichess4545leaguerulesregulations",
+                "team": "https://lichess4545.slack.com/files/mrlegilimens/F0VNACY64/lichess4545season3-graphs",
+                "guide": "http://bit.ly/1MW9YJ4",
+                "captains": "https://lichess4545.slack.com/files/endrawes0/F0V3SPE90/guidelinesforlichess4545teamcaptains2.doc",
+                "registration": "https://docs.google.com/a/georgetown.edu/forms/d/1u-fjOm1Mouz8J7WAsPhB1CJpB3k10FSp4-fZ-bwvykY/viewform"
+            }
+        },
+        "lonewolf": {
+            // TODO:
```
endrawes0: what is todo?

fuzzy_match.js
```diff
@@ -47,13 +42,34 @@ function rank_choices(search_string, targets) {
...
+
+        // You have to get it at least half right
+        if (distance < searchString.length/2) {
+            results.push([distance, item]);
+        }
+    });
+    return findBestMatches(results, includeDistance);
+}
+
+// TODO: I don't believe this is used anymore, if so rip it out
```
endrawes0: could create an issue. idc.

config.js
lakinwecker: Nothing anymore. I will remove.

config.js
lakinwecker: Thanks!  I'm happy with how it turned out as well.

http.js
```diff
@@ -0,0 +1,41 @@
...
+        res.on('end', () => {
+            deferred.resolve({'response': res, "body": body});
+        });
+    }).on('error', (e) => {
+        console.error(JSON.stringify(e));
+        deferred.reject("failed to get a response from url: " + url);
+    });
+    return deferred.promise;
+}
+function fetchURLIntoJSON(url){
```
endrawes0: if we ever need a new version of this function, say `fetchURLIntoXML` or something, we should just make it driven by an encoding.

`fetchURL(url, [encoding])` where encoding is text, json, XML etc.

Just food for thought. This seems more in line with similar apis.

league.js
```diff
@@ -0,0 +1,693 @@
+//------------------------------------------------------------------------------
+// Defines a league object which can be used to interact with the spreadsheet
+// for the given league
+//------------------------------------------------------------------------------
+// TODO: create a set of slack helpers that map channel names to a
+// specific league object, so that the code for scheduling, results or
+// gamelinks
```
endrawes0: you did not finish your thought here.

league.js
```diff
@@ -0,0 +1,693 @@
...
+                console.error("Unable to refresh schedule: " + err);
+                throw new Error(err);
+            } else {
+                console.log("Found " + pairings.length + " pairings for " + self.options.name);
+            }
+            self._lastUpdated = moment.utc();
+        });
+    },
+
+    //--------------------------------------------------------------------------
```
endrawes0: will pick up  from here tomorrow or later. thanks @lakinwecker

chesster.js
endrawes0: me neither. just curious. dont touch it.

chesster.js
endrawes0: I recall, regarding the handler's inability to deal with failures in asynch contexts.

league.js
lakinwecker: Thanks! The TODO is no longer valid anyways, so I'll just remove it.

http.js
lakinwecker: Agreed! That would be nice.

chesster.js
endrawes0: Yeah. that is very nearly what I was thinking... Exact interaction TBD. But that is the idea I have in mind.

chesster.js
endrawes0: definitely can go in an issue. #58

I am too lazy to create issues now. Will mark stuff that needs an issue with #58 and come back to it tomorrow.

fuzzy_match.js
endrawes0: #58

chesster.js
endrawes0: #58

chesster.js
endrawes0: #58

league.js
```diff
@@ -0,0 +1,693 @@
...
+                    if (err != "Unable to find target worksheet") {
+                        return callback(err, rows);
+                    } else {
+                        return callback(undefined, []);
+                    }
+                }
+                var newTeams = [];
+                rows.forEach(function(row) {
+                    if (
+                        !row['teams'].value ||
```
endrawes0: I dont really like that the number of boards is hard coded - this may need to be more flexible. Fine for now but #58 to make this configurable some how.

league.js
```diff
@@ -0,0 +1,693 @@
...
+            'min-col': 1,
+            'max-col': 8,
+            'return-empty': true
+        }
+        var self = this;
+        spreadsheets.getPairingRows(
+            self.options.spreadsheet,
+            query_options,
+            function(err, rows) {
+                if (err) { return callback(err, rows); }
```
endrawes0: this call back takes rows and the callback below takes pairings.

should this argument be left off?

league.js
```diff
@@ -0,0 +1,693 @@
...
+        });
+    },
+    //--------------------------------------------------------------------------
+    // Formats the captains Guidelines
+    //--------------------------------------------------------------------------
+    'formatCaptainGuidelinesResponse': function() {
+        var self = this;
+        return Q.fcall(function() {
+            if (self.options.links && self.options.links.captains) {
+                return "Here are the captain's guidelines:\n" + self.options.links.captains;
```
endrawes0: #58 we should be consistent about how we format strings.

lichess.js
```diff
@@ -0,0 +1,143 @@
...
+            var isJSON = request[1];
+            var deferred = request[2];
+            var promise = null;
+            if (isJSON)  {
+                promise = http.fetchURLIntoJSON(url);
+            } else {
+                promise = http.fetchURL(url);
+            }
+            promise.then(function(result) {
+                if (result.response.statusCode == 429) {
```
endrawes0: in case of a 429 in response to a user message, how much effort would it be to let the user know? is this information accessible from from that context? Just a potential enhancement. #58

lichess.js
```diff
@@ -0,0 +1,143 @@
...
+            });
+        } else {
+            setTimeout(processRequest, requestDelay);
+        }
+    }
+    processRequest();
+    return makeRequest;
+})();
+
+// TODO: eventually cache these results.
```
endrawes0: TODO is TODONE? :)

slack.js
```diff
@@ -0,0 +1,363 @@
...
+        _.each(league.getAllLeagues(config), function(l) {
+            l.refresh();
+        });
+        setTimeout(function(){ 
+            refresh(bot, delay, config);
+        }, delay);
+    }));
+}
+
+function criticalPath(promise){
```
endrawes0: #58 we should refactor into an exception handler module.

slack.js
```diff
@@ -0,0 +1,363 @@
...
+                    console.error("Middleware asked to not process controller callback: " + JSON.stringify(error));
+                } else {
+                    throw error;
+                }
+            });
+
+        }));
+    });
+}
+DEFAULT_HEARS_OPTIONS = {
```
endrawes0: code so nice you wrote it twice.

slack.js
```diff
@@ -0,0 +1,363 @@
...
+    middleware: []
+};
+function hears(options, callback) {
+    var self = this;
+    var options = _.extend({}, DEFAULT_HEARS_OPTIONS, options);
+    self.controller.hears(options.patterns, options.messageTypes, function(bot, message) {
+        return botExceptionHandler(bot, message, Q.fcall(function() {
+            message.player = users.getByNameOrID(message.user);
+            // This will occur if a new player uses a chesster command
```
endrawes0: this could likely happen if the welcome becomes interactive. just something to keep in mind.

test/test_league.js
```diff
@@ -0,0 +1,751 @@
...
+    }
+}
+if (private_key) {
+    _45_45_LEAGUE_CONF["spreadsheet"]["serviceAccountAuth"] = {
+        "client_email": "tesster@chesster-lichess-4545-bot.iam.gserviceaccount.com",
+        "private_key": private_key
+    }
+}
+
+describe('league', function() {
```
endrawes0: ill admit, I am not looking closely at the tests.

endrawes0:
I am happy with everything here - there are a few minor things to clean up - comments, some code. Let me know when you are ready to merge. Im going to create some issues and say my job is done.

endrawes0:
Issues are created.

Minor stuff then merge!

Great work @lakinwecker

league.js
lakinwecker: Good point. Fixed.

league.js
lakinwecker: I added an issue for this.

lichess.js
lakinwecker: TODONE!

slack.js
lakinwecker: Agreed.

slack.js
lakinwecker: Fixed.

PR #41, opened 2016-05-15
Some defense

endrawes0:
In this pull request:

1. replaced error_log with stderr.
2. fetching json from a url will now return the json object, potenially undefined, and an error. I wanted to do this using exceptions, but because of the callbacks, that was not possible.
3. where the function is used, I check the result of the json and if not defined, then write the error to the logs and handle the error.
4. added color to pairings.
5. I refectactored some of the pairing code while I was handling the errors and colors.

### DIFF
--- a/chesster.js
+++ b/chesster.js
@@ -41,16 +41,8 @@ function exception_handler(todo, on_error){
             "\nDatetime: " + new Date() +
             "\nError: " + JSON.stringify(e) +
             "\nStack: " + e.stack;
-        console.log(error_log);
-        fs.appendFile("./errors_log", error_log, function(err) {
-            if(err) {
-                console.log("failed to write to the file...")
-                console.log(e);
-                return console.log(err);
-            }
-        });
+        console.error(error_log);
         on_error && on_error();
-        console.log("\n");
     }
 }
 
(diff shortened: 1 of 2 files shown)


chesster.js
```diff
@@ -698,28 +691,58 @@ function parsePairingForRound(roundSheet, requestingPlayer, targetPlayer, callba
...
+                    "opponent": opponentName, 
+                    "color": color, 
+                    "date": date, 
+                    "time": time, 
+                    "rating": rating, 
+                }, callback);
+            });
+        }else{
+            console.error(JSON.stringify(error));
+            formatPairingResult({
```
lakinwecker: Seems odd that you are calling this function in the error case as well?

endrawes0: in the case that we cannot find a rating, it could be because the api is down. 

no reason to not display the pairing without the rating.

lakinwecker: :+1: I missed that when I read it the first time.

aweidner:
Typically the node style callback is `function(error, result)`.  I'd prefer if we kept that convention as long as there is no strong reason for `function(result, error)`.

endrawes0:
Seems odd to me, but ok. I'll make the change.

endrawes0:
reversed the order of the arguments for the callback on feth_json_from_url

going to merge now.

PR #42, opened 2016-05-15
previously, this created a separate log for stderr to 1. this is what…

endrawes0:
… was originally inteneded.


endrawes0:
minor change - not requesting review.

PR #43, opened 2016-05-17
Add travis.yml file

aweidner:
This isn't ready for merge yet, just want to get some discussion going on this.  The steps for travis are basically simple.  Somebody (most likely @endrawes0) needs to sign up with travis and have admin (?) access to the Chesster repository.  Then we just add this .travis.yml file and all pull requests, branches, etc will be built automatically.

The integration tests run fine on Travis, so unless there is a reason not to run them, we can run them as part of our standard build.

In order to allow travis to download dependencies, I also added a default package.json with our current  deps.  Right now it's set to download the latest, but we can lock them to specific versions as well (I think we discussed that before).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,4 @@
+language: node_js
+node_js: "5"
+install: npm install
+script: mocha
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,4 @@
 language: node_js
 node_js: "5"
 install: npm install
-script: mocha
+script: mocha --timeout 5000
--- a/package.json
+++ b/package.json
@@ -1,0 +1,34 @@
+{
+  "name": "chesster",
+  "version": "1.0.0",
+  "description": "Bot for Lichess4545 Slack group",
+  "main": "chesster.js",
+  "directories": {
+    "test": "test"
+  },
+  "dependencies": {
+    "async": "^2.0.0-rc.3",
+    "botkit": "^0.1.1",
+    "chai": "^3.5.0",
+    "fast-levenshtein": "^1.1.3",
+    "google-spreadsheet": "^2.0.1",
+    "moment": "^2.13.0",
+    "underscore": "^1.8.3"
+  },
+  "devDependencies": {
+    "mocha": "^2.4.5"
+  },
+  "scripts": {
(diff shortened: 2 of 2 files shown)


.travis.yml
```diff
@@ -0,0 +1,4 @@
+language: node_js
```
endrawes0: I am terribly uneducated here. Can you describe this file, how it is used and what each of lines represents?

Thanks!

package.json
```diff
@@ -0,0 +1,34 @@
+{
+  "name": "chesster",
+  "version": "1.0.0",
```
endrawes0: we are officially 1.0? woohoo! we made it out of alpha and beta without even knowing it!

package.json
```diff
@@ -0,0 +1,34 @@
...
+  },
+  "devDependencies": {
+    "mocha": "^2.4.5"
+  },
+  "scripts": {
+    "test": "mocha"
+  },
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/aweidner/Chesster.git"
```
endrawes0: these should point to my repo, I presume.

.travis.yml
aweidner: Ah yeah my bad
1. Set the language of the project to node_js
2. Set the current version of the nodejs instance running on the travis testing container to the latest version of the 5.x branch of nodejs.  We can set specific versions if we need to, but for now I think latest should be alright.
3.  When we do the install step, run `npm install` to install dependencies.  Commands in the install section are always run before commands in the script section, so we will have our dependencies installed for our test
4. Run tests using the mocha test runner.  Set the timeout of the tests to 5000 milliseconds (for the integration tests).  The default time is 2000 ms, which we could potentially hit with a valid request.

package.json
aweidner: Yeah correct, and the version should be whatever the current version of Chesster is.  I'm not sure, do we have it somewhere?

.travis.yml
endrawes0: Thank you for the careful explanations!

package.json
endrawes0: we have never needed a version number before.

how about 42?

PR #48, opened 2016-05-19
Removes changelog

aweidner:


endrawes0:
Thanks for this. There is actually some code in Chesster that will also need cleaned up.

endrawes0:
I was not suggesting you should do it - unless you wanted to. Just pointing it out.

aweidner:
I think I got everything that needed to be removed from this, I merged it up with master too.  Is there anything else that needs to be done?

lakinwecker:
I merged it with my long-standing branch. So it's actually merged with master via that branch

lakinwecker:
Thanks!

PR #52, opened 2016-06-03
Updating Mod List

krcarriere:
Just added the three new mods un prepareModsMessage() and prepareSummonModsMessage() . Was there any other place I needed to update?

### DIFF
--- a/chesster.js
+++ b/chesster.js
@@ -417,7 +417,7 @@ function prepareSummonLoneWolfModsMessage(){
  * works fine.
  */
 function prepareModsMessage(){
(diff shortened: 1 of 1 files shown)


chesster.js
```diff
@@ -414,7 +417,7 @@ function prepareSummonLoneWolfModsMessage(){
  * works fine.
  */
 function prepareModsMessage(){
-    return "Mods: endrawes0, mkoga, mrlegilimens, petruchio, seb32, t\u200Bheino";
+    return "Mods: endrawes0, mkoga, mrlegilimens, petruchio, seb32, prune2000, angborxley, prune2000, t\u200Bheino";
```
lakinwecker: You added prune2000 twice and skipped tnan123 here.

PR #73, opened 2016-06-17
Removes unused match function from fuzzy_matcher

aweidner:
Confirmed that this function is not being used anywhere in the current code base.  We can safely remove it.


endrawes0:
Looks good.

endrawes0:
Thanks!

PR #75, opened 2016-06-18
Adds ability to generate code coverage for Code Climate

aweidner:


endrawes0:
still working on this right?

aweidner:
Yep, still need to think of a clean way to do this, ideally without some script or other hack around this.

endrawes0:
The tests were modified because updated the spreadsheet with a stalemate for testing last night.

aweidner:
Okay, so the way I'm doing this is to basically ignore failure.  There's a section in the .travis.yml file that runs tasks after the main task has completed.  Those tasks are not taken into account for the status of the main build.  Ideally, I would be using that but I wasn't able to get it to actually send the code coverage information.

Instead, I added a npm script entry in the package.json file to run the code coverage command and ignore failure (exit 0 always).

endrawes0:
By ignore failure, you mean exit with status 0 so mocha rubs first and regardless of whether or not coverage succeeds, the entire ci passes.

So we can't RELY on travis for the result of the coverage but rather the fact that it will run.

We will need to check the code coverage results separately.

Did i understand correctly?

aweidner:
Yeah, mocha is the only task which fails the build.  If a build passes, code coverage is not 100% guaranteed to have received coverage.  Coverage will always be generated (as a result of running mocha), but it's not guaranteed to have been sent.  Also there's about a 20 minute delay between code climate receiving coverage and it being displayed.

endrawes0:
Ok. So, push, wait for mocha to finish. Wait 20, check code coverage - if necessary. Merge.

aweidner:
Yep, and if we want to require a certain amount of code coverage, we can do that through mocha.

endrawes0:
I am content with these changes. 

Anything else before I merge?

aweidner:
Nope, this should do it

PR #84, opened 2016-06-25
Add linting

aweidner:
Still need to test this to verify all the changes did not break anything, but this changeset will allow us to add eslint as part of our travis builds.  We will need to configure code climate separately.


src/lichess.js
```diff
@@ -75,7 +75,7 @@ var makeRequest = (function() {
     }
     processRequest();
     return makeRequest;
-})();
+}());
```
lakinwecker: I don't understand why this version is preferable to the other one.

src/slack.js
```diff
@@ -164,7 +164,7 @@ function isModerator(message) {
 //------------------------------------------------------------------------------
 function requiresModerator(bot, message, config) {
     return Q.fcall(function() {
-        if (typeof message.league == 'undefined') {
+        if (_.isUndefined(message.league)) {
```
lakinwecker: oooh. Nice. I much prefer this.

endrawes0:
PR looks  good. after testing, we can merge. Are we still planning to replace tests for string equality with Lo-Dash `isEqual`?

src/lichess.js
aweidner: Looking at the rule, it seems that either is the same and this is purely a stylistic choice.  http://eslint.org/docs/2.0.0/rules/wrap-iife.  We can set it to inside, outside, or any.  I can change the rule and revert it if you would rather have outside.  I don't have a preference either way.

aweidner:
Yes, I will be changing anything were string comparisons from an unknown source (one of the third party libraries or user input) are involved to `isEqual`.

aweidner:
One thing to note about the change to `_.isEqual`.  It is used only in the case where both the comparisons are expected to be against string data, and both of the operands are unknown.  In the case where we are comparing two integers, `===` should serve us fine.  Also in the case where we are comparing a string against a known value (something like `playerName === "ghostologist"`), `===` will work also.  The only time we need to use `_.isEqual` is in the case where we have `playerName === otherName` as we don't know either operand, and we could encounter the `new String("") !== ""` case.

The only exception to this rule is where we are comparing explicitly against `""` in which case we must also use`_.isEqual`.

src/chesster.js
```diff
@@ -587,7 +588,7 @@ function(bot, message) {
         results,
         function(err, reversed) {
             if (err) {
-                if (err.indexOf && err.indexOf("Unable to find pairing.") == 0) {
+                if (_.includes(err, "Unable to find pairing.")) {
```
lakinwecker: This is great! :)

src/chesster.js
```diff
@@ -830,22 +831,22 @@ function validateUserResult(details, result){
         valid: true,
         reason: ""
     };
-    if( details.winner && result.result == "1/2-1/2" ){
+    if( details.winner && result.result === "1/2-1/2" ){
```
lakinwecker: This is probably fine - but this is a case where I think _.isEqual is probably better.

src/league.js
```diff
@@ -529,15 +528,15 @@ league_attributes = {
...
         return Q.fcall(function() {
-            if (self._teams.length == 0) {
+            if (self._teams.length === 0) {
                 return "The {name} league does not have captains".format({
                     name: self.options.name
                 });
             }
             teams = _.filter(self._teams, function(t) {
-                return t.name.toLowerCase() == teamName.toLowerCase()
+                return t.name.toLowerCase() === teamName.toLowerCase()
```
lakinwecker: This is the same case as below, we should use _.isEquals for both or neither. I suspect that toLowerCase() is pretty consistent about what types it returns, so === is probably fine.

src/lichess.js
lakinwecker: Ok.  I don't have a stylistic preference either. Let's leave it the default for eslint. Easier that way.

lakinwecker:
This is great stuff. I just reviewed it again and I'm really happy with it.

Looking forward to seeing this go live.

endrawes0:
I am ready to merge this. Any concerns?

src/league.js
```diff
@@ -281,10 +283,7 @@ league_attributes = {
     // Returns whether someone is a moderator or not.
     //--------------------------------------------------------------------------
     'isModerator': function(name) {
-        var self = this;
-        return _.some(self.options.moderators, function(moderator) {
-            return name == moderator;
-        });
+        return _.includes(this.options.moderator, name);
```
lakinwecker: This should be plural: ```this.options.moderators```

endrawes0: Fixing this in my PR. #86

PR #85, opened 2016-06-26
Adds a local database and uses it to cache ratings

lakinwecker:
This commit adds a local sqlite3 database and uses said database to cache player ratings.

This is ready for comments. The primary focus should be on the database integration.  Some notes:

1. I had to write our own locking mechanism to ensure that all queries are run to completion before a new one is run. This will slow us down. We should be able to turn on the WAL and ONLY run writes through this locking mechanism moving forward.
2. I'm not a big fan of how I've had to organize the config and various files - but I'm not sure we can do any better, considering. 
3. sequelize-cli is required to run migrations and make migrations etc. I've been running it locally like so: ```./node_modules/sequelize-cli/bin/sequelize --config ./config/config.js db:migrate``` but we'll want to figure something else out for that. I think we can add custom commands to our package.json, so maybe we can do it that way.

### DIFF
--- a/config/config.js
+++ b/config/config.js
@@ -4,6 +4,19 @@ var token = require("./slack_token.js").token;
 var private_key = require("./service_account_key.js").key;
 
 var config = {
+    // Unfortunately this all has to be at the top level due to sequelize-cli
+    "database": "",
+    "username": "",
+    "password": "",
+    "dialect": "sqlite",
+    "logging": false,
+	"pool": {
+		"max": 5,
+		"min": 0,
+		"idle": 10000
+	},
+    "storage": "./db/database.sqlite",
+
     "token": token,
     "links": {
         "source": "https://github.com/endrawes0/Chesster"
@@ -150,4 +163,4 @@ var config = {
         "unstable_bot": "45+45"
     },
 }
-module.exports.config = config;
+module.exports = config;
--- a/db/empty.txt
+++ b/db/empty.txt
@@ -1,0 +1,1 @@
+So we can commit the directory, but ignore the db inside of it.
--- a/migrations/20160625024216-adding_rating_table.js
+++ b/migrations/20160625024216-adding_rating_table.js
@@ -1,0 +1,45 @@
+'use strict';
+
+module.exports = {
+    up: function (queryInterface, Sequelize) {
+        /*
+            Add altering commands here.
+            Return a promise to correctly handle asynchronicity.
+
+            Example:
+            return queryInterface.createTable('users', { id: Sequelize.INTEGER });
+        */
+        return queryInterface.createTable('LichessRatings', {
+            id: {
+                type: Sequelize.INTEGER,
+                autoIncrement: true,
+                primaryKey: true
+            },
+            lichessUserName: Sequelize.STRING, // Comes from lichess
+            rating: {
+                type: Sequelize.INTEGER, allowNull: true
+            },
+            lastCheckedAt: {
+                type: Sequelize.DATE, allowNull: true
+            },
+            // These two are automatically added to a model by sequelize
+            createdAt: {
+                type: Sequelize.DATE, allowNull: true
+            },
+            updatedAt: {
+                type: Sequelize.DATE, allowNull: true
+            }
+        });
+    },
+
+    down: function (queryInterface, Sequelize) {
+        /*
+            Add reverting commands here.
+            Return a promise to correctly handle asynchronicity.
+
+            Example:
+            return queryInterface.dropTable('users');
+        */
+        return queryInterface.dropTable('LichessRatings');
+    }
+};
--- a/src/lichess.js
+++ b/src/lichess.js
@@ -4,6 +4,10 @@
 var http = require("./http.js");
 var moment = require("moment");
 var Q = require("q");
+// TODO: sequelize-cli requires us to call this models.js or models/index.js
+//       this name conflicts with the parameter that we pass in after getting
+//       the lock, so I'd like to (by convention) always refer to this as db.
+var db = require("./models.js");
 
 var MILISECOND = 1;
 var SECONDS = 1000 * MILISECOND;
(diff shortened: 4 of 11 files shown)


endrawes0:
I'm not familiar with migrations wrt databases, can you explain this concept to me?

As for custom commands, i believe you are correct, so that idea is sound. 

Will review today.

lakinwecker:
Migrations are a set of ordered files that contain the necessary commands to alter a database into the required 'current' form.  In this case, there is only one migration which creates the necessary LichessRatings table.  However, as we move forward there will be PRs with new tables or new columns or changed columns and we want to have a single command to properly takes the DB from whatever state it's currently in and changes it to the new state without losing data.

Migrations accomplish this by keeping a log of which migrations have been applied to a particular database in the database itself. They compare this to the list of available migrations and then run the ones that have not yet been run.

config/config.js
```diff
@@ -4,6 +4,19 @@ var token = require("./slack_token.js").token;
 var private_key = require("./service_account_key.js").key;
 
 var config = {
+    // Unfortunately this all has to be at the top level due to sequelize-cli
```
endrawes0: We can at least move it to the bottom of the config so it's not the most prominent part of the config.

config/config.js
```diff
@@ -150,4 +163,4 @@ var config = {
         "unstable_bot": "45+45"
     },
 }
-module.exports.config = config;
```
endrawes0: I didn't realize you could assign directly to exports. That's better.

db/empty.txt
```diff
@@ -0,0 +1 @@
+So we can commit the directory, but ignore the db inside of it.
```
endrawes0: Maybe some day we will actually need something here ACS this can go away.

Perhaps a README for setting up the db.

migrations/20160625024216-adding_rating_table.js
```diff
@@ -0,0 +1,45 @@
...
+module.exports = {
+    up: function (queryInterface, Sequelize) {
+        /*
+            Add altering commands here.
+            Return a promise to correctly handle asynchronicity.
+
+            Example:
+            return queryInterface.createTable('users', { id: Sequelize.INTEGER });
+        */
+        return queryInterface.createTable('LichessRatings', {
```
endrawes0: Schema definition is interesting this way. Only ever defined or in SQL.

lakinwecker: The idea behind an ORM like this is to be able to easily switch between DB types. Of course, it's never perfectly replaceable in practice, but for projects like these - it's close.

src/lichess.js
```diff
@@ -91,52 +95,93 @@ function getPlayerByName(name, isBackground){
...
+                    // If we don't have a rating, use the foreground queue
+                    promise = _updateRating(name, false);
+                } else if (!isInQueue && (!lastCheckedAt || lastCheckedAt.isBefore(_30MinsAgo))) {
+                    // If the rating is just out of date, use the background queue
+                    promise = _updateRating(name, true);
+                }
+
+                if (rating !== null && typeof rating !== 'undefined')  {
+                    // Return the cached rating if we have it.
+                    return lichessRating.get('rating');
```
endrawes0: All results are promises, correct? There should be no ambiguity in the return value of this function.

Based on usage in league.js, this appears to be the case, just confirming.

src/models.js
```diff
@@ -0,0 +1,76 @@
...
+            var unlock = Q.defer();
+            lockDeferred.resolve(unlock);
+            return unlock.promise;
+        });
+        return lockDeferred.promise;
+    }
+
+    //--------------------------------------------------------------------------
+    function defineModels(sequelize) {
+        module.exports.LichessRating = sequelize.define('LichessRating', {
```
endrawes0: How/why does the definition of this model differ from the database migrations? It seems the model does not expose creation and update details. 

Will a separate model contain this information? Some audit model?

src/models.js
```diff
@@ -0,0 +1,76 @@
...
+    // Connection function that ensures we have a connection and the models
+    // are defined.
+    //
+    // Parameters: config - the config option that contains the database
+    //                      information.
+    //--------------------------------------------------------------------------
+    function connect(config) {
+        var sequelize = new Sequelize(config.database, config.username, config.password, config);
+        var deferred = Q.defer();
+        return sequelize.authenticate().then(function() {
```
endrawes0: 2 return values?

I assume this is a typo. The correct return value is the deferred, yeah?

src/models.js
```diff
@@ -0,0 +1,76 @@
...
+            defineModels(sequelize);
+            deferred.resolve();
+        }).catch(function(error) {
+            console.error("Unable to connect to database: " + error);
+            deferred.reject(error);
+        });
+        return deferred.promise;
```
endrawes0: Unreachable

endrawes0:
A few minor issues to address / questions to answer.

Moving onto new PR.

src/models.js
lakinwecker: I assume you saw the comment about sequelize adding the two extra columns by now.  That's one way.

Also, these models will be different from the migrations if/when we add remove or change columns.  In that case, a new migration would be made and ONLY the change would be inserted into that migration.  At that point, this model definition would represent the current definition, and the migrations would represent the original definition followed by a series of changes that would modify the original definition into the current definition.

src/models.js
lakinwecker: Fixed in my other branch.

src/models.js
lakinwecker: Fixed in my other branch.

src/lichess.js
lakinwecker: I'm sorry, I don't understand your question here?  Promises are return until the end of the chain when a final "value" is returned. This is ending the chain by returning the final value.

Does that answer your question?

src/models.js
endrawes0: Well said. That explains it.

My main concern was that definitions would be redundant. You have made it more clear, that migrations are historical in nature. 

Thanks

src/lichess.js
endrawes0: No but no need. I answered it myself. The question was not stated well.

I really just wanted to confirm the return value of `LichessRating.get`

All set. Thanks.

Issue #83, opened 2016-06-25
Some commands trigger 3 error messages


lakinwecker:
lonewolf 45+45 this is a long sentence, maybe it requires a long sentence, and some numbers 24 2

PR #86, opened 2016-06-26
Vagrant and league recognition fix

endrawes0:
The issue is, we were only handling errors of the particular type: `StopControllerError`. The error was of type `Error` and thus went unhandled.

Also, we were passing an error message but it was never stored so dumping that error to the logs was just ignored.

I have saved it as a member of `StopControllerError`.

Updated Vagrantfile to include a script to run on startup - to acquire the most recent versions of our dependencies.

Lastly, update the location of the vagrant-dependencies.sh to be run on first `vagrant up` after a `vagrant destroy`.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -37,6 +37,6 @@ Vagrant.configure(2) do |config|
   config.vm.hostname = "chesster"
 
 
-  config.vm.provision :shell, :path => "./setup/vagrant-dependencies.sh"
-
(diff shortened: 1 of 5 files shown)


endrawes0:
this addresses #83

lakinwecker:
Looks great to me.

endrawes0:
Thanks -going to do a release

PR #87, opened 2016-06-27
Event subscription system

lakinwecker:
WIP - this is the initial version of a working event subscription system. Please note that this branch is based off of my other PR work for the models so the models stuff is also shown here. Once that is merged, this will be a bit more lean. Lots to do here still - but time to start the review.


config/config.js
```diff
@@ -4,6 +4,19 @@ var token = require("./slack_token.js").token;
 var private_key = require("./service_account_key.js").key;
 
 var config = {
+    // Unfortunately this all has to be at the top level due to sequelize-cli
```
endrawes0: Looking at how this is used, I don't see why this can't be organized in an inner object that gets passed to sequelize.

Can you detail this limitation further?

migrations/20160625024216-adding_rating_table.js
```diff
@@ -0,0 +1,45 @@
...
+                primaryKey: true
+            },
+            lichessUserName: Sequelize.STRING, // Comes from lichess
+            rating: {
+                type: Sequelize.INTEGER, allowNull: true
+            },
+            lastCheckedAt: {
+                type: Sequelize.DATE, allowNull: true
+            },
+            // These two are automatically added to a model by sequelize
```
endrawes0: Ah. I see. Didn't notice this comment before.

migrations/20160626215331-adding_subscription_table.js
```diff
@@ -0,0 +1,71 @@
...
+    up: function (queryInterface, Sequelize) {
+        /*
+            Add altering commands here.
+            Return a promise to correctly handle asynchronicity.
+
+            Example:
+            return queryInterface.createTable('users', { id: Sequelize.INTEGER });
+        */
+        return Q.all([
+			queryInterface.createTable('Subscriptions', {
```
endrawes0: Indentation looks off her

src/chesster.js
```diff
@@ -970,3 +972,87 @@ function(bot, message) {
...
+            deferred.resolve();
+        }).catch(function(error) {
+            convo.say("I'm sorry, but an error occurred processing this subscription command");
+            deferred.reject(error);
+        });
+    });
+    return deferred.promise;
+});
+
+subscription.emitter.on('a-game-is-scheduled', function(bot, message, l, results, white, black) {
```
endrawes0: Please change `l` to `league`.

src/models.js
```diff
@@ -0,0 +1,98 @@
...
+
+    //--------------------------------------------------------------------------
+    // Connection function that ensures we have a connection and the models
+    // are defined.
+    //
+    // Parameters: config - the config option that contains the database
+    //                      information.
+    //--------------------------------------------------------------------------
+    function connect(config) {
+        var sequelize = new Sequelize(config.database, config.username, config.password, config);
```
endrawes0: Why can't this be buried in the config?

Seems 
`Sequelize(config.db.database, ..., config.db)` should be equivalent.

Is sequelize using something at the top level of config?

src/subscription.js
```diff
@@ -0,0 +1,278 @@
...
+        var tell = args[0],
+            listener = args[1],
+            when = args[2],
+            event = args[3],
+            _in = args[4], 
+            targetLeague = args[5], 
+            _for = args[6];
+
+        // Ensure the basic command format is valid
+        if (!_.isEqual(["tell", "when", "in", "for"], [tell, when, _in, _for])) {
```
endrawes0: Very explicit. Well done.

src/subscription.js
```diff
@@ -0,0 +1,278 @@
...
+                return "Great! I will tell {target} when {event} for {source} in {league}".format({
+                    source: source.name,
+                    event: event,
+                    league: l.options.name,
+                    target: listener
+                });
+            });
+        });
+    });
+}
```
endrawes0: Looks like the basics are in place. Extension for teams and targets should be straightforward.

src/subscription.js
```diff
@@ -0,0 +1,278 @@
...
+                    unlock.resolve();
+                    throw Error("This should never occur");
+                }
+            });
+        });
+    });
+}
+
+//------------------------------------------------------------------------------
+// Get listeners for are a given event and source
```
endrawes0: ~are~

endrawes0:
I cannot unfortunately spend more time on this until later. I'll bring slack up at the office. Let's chat more there.

Great work. Your productivity is bewildering. This would have taken me twice as long.

Thanks!

config/config.js
lakinwecker: sequelize-cli is the command line tool we use to apply and generate migrations. It looks for it at the top level in the config.  :( 

I could, however, move it to a config/database.js or some such and have sequelize-cli use it from there, and import it into the db namespace in here.

migrations/20160626215331-adding_subscription_table.js
lakinwecker: Thanks, Fixed.

src/chesster.js
lakinwecker: The issue here is that 'league' is already taken in the top of this file by the var league = require("./league.js");

I was trying to ensure we were not conflicting.  I can change it, but I'm not certain enough about the scoping rules of javascript to know if that will cause issues or not.

src/models.js
lakinwecker: as I commented above, sequelize-cli is the real cause.

src/chesster.js
endrawes0: Oh I see.

I prefer using `let` instead of `var`. It makes scope more sensible.

Our version of node doesn't support it however.

src/models.js
endrawes0: Understood. Thanks.

src/chesster.js
```diff
@@ -970,3 +972,87 @@ function(bot, message) {
...
+            deferred.resolve();
+        }).catch(function(error) {
+            convo.say("I'm sorry, but an error occurred processing this subscription command");
+            deferred.reject(error);
+        });
+    });
+    return deferred.promise;
+});
+
+chesster.hears({
```
aweidner: One thing that might be convenient is to have the subscription command work the same way git does in regards to syntax:

subscription status - Show subscriptions
subscription remove (sub) - Removes a subscription
subscription add (event) - Adds a subscription

Obviously this is just a suggestion, but it might make it more straightforward to add new commands in the future if that's necessary as we'll have a standard model for how the commands are structured.

src/chesster.js
```diff
@@ -970,3 +972,87 @@ function(bot, message) {
...
+                    bot.startPrivateConversation({user: target.id}, function(err, convo) {
+                        // TODO: this could be a much better message!
+                        convo.say("{white} vs {black} has been scheduled for {date}.".format({
+                            white: white.name,
+                            black: black.name,
+                            date: "TODO"
+                        }));
+                    });
+                } else {
+                    // TODO: probably delete the subscription at this point.
```
aweidner: Yeah how would this get triggered exactly?  The user no longer exists in slack?  If that's the case, then the subscription should be deleted as this comment indicates.

src/lichess.js
```diff
@@ -91,52 +95,93 @@ function getPlayerByName(name, isBackground){
...
+                    lastCheckedAt = moment.utc(lastCheckedAt);
+                }
+                var promise;
+                var isInQueue = _playerIsInQueue[name];
+                var rating = lichessRating.get('rating');
+
+                // Only update the rating if it's older than 30 minutes
+                // or if we don't have one a rating
+                // TODO: Replace this with _.isInteger
+                if (rating === null || typeof rating === 'undefined')  {
```
aweidner: When we switch to lodash, this is better written as `_.isNil`.  Just making this comment to mark the place so it can be replaced later.

lakinwecker: Cool. that would be better.

src/chesster.js
lakinwecker: That's correct. If the user is no longer part of slack. I agree, it should be deleted.

endrawes0: That was my interpretation

src/chesster.js
lakinwecker: Good idea. I like it.  I will switch these two commands to using that format. I will use subscription list
subscription remove (id).
I still like the tell version of the command as I think that's easier to remember and understand for non technical users.

src/player.js
```diff
@@ -25,6 +27,16 @@ module.exports.getSlackUser = function(users, message) {
     if (message.match[1]) {
         nameOrId = message.match[1];
     }
+    var player = getSlackUserFromNameOrID(users, nameOrId);
```
aweidner: I think this can be written

`var player = getSlackUserFromNameOrID(users, nameOrId) || users.getByNameOrID(message.user)`

lakinwecker: :+1: 

Also, I the getPlayer method should probably exist in src/slack.js and the end users shouldn't have to pass in slack.users.  It should just use that directly. 

Thoughts?

src/subscription.js
```diff
@@ -0,0 +1,278 @@
...
+    "a-pairing-is-released"
+];
+
+//------------------------------------------------------------------------------
+// Format the help response message
+//------------------------------------------------------------------------------
+function formatHelpResponse(config) {
+    return Q.fcall(function() {
+        var leagueNames = [];
+        _.each(league.getAllLeagues(config), function(l) {
```
aweidner: Can we refactor the segment for getting league names into a separate function?

```
function getLeaugeNames() {
    return _.map(leauge.getAllLeauges(config), function(l) {
        return l.options.name
    }
}
```

src/subscription.js
aweidner: Yeah this is very good.  I sitting here asking "what's magic about the number 7 that makes this work?", but the variable names explain it beautifully.

src/subscription.js
lakinwecker: :+1:

src/subscription.js
lakinwecker: I really wish javascript had this from python:
```tell, listener, when, event, _in, targetLeague, _for  = components.slice(0,7)```

this was the next best thing.

File: src/subscription.js [hunk 14]
```diff
@@ -0,0 +1,278 @@
...
+                    return "That is not a valid subscription id";
+                } else if (subscriptions.length === 1) {
+                    return subscriptions[0].destroy().then(function() {
+                        unlock.resolve();
+                        return "Subscription deleted";
+                    });
+                } else {
+                    // This should never happen
+                    unlock.resolve();
+                    throw Error("This should never occur");
```
aweidner: May want a more explicit error message here describing exactly what happened.  It's quite easy for someone to grep the code for "This should never occur" and read the surrounding code to get the context, but it might save somebody 5 minutes if the impossible happens someday.

src/subscription.js
lakinwecker: In python it's even more elegant:
 tell, listener, when, event, _in, targetLeague, _for = message.text.split(" ", 7)

src/subscription.js
```diff
@@ -0,0 +1,278 @@
...
+    return db.lock().then(function(unlock) {
+        return db.Subscription.findAll({
+            where: {
+                league: leagueName.toLowerCase(),
+                source: source.toLowerCase(),
+                event: event.toLowerCase()
+            }
+        }).then(function(subscriptions) {
+            unlock.resolve();
+            var listeners = [];
```
aweidner: one more for the lodash refactor pile:

`return _(subscriptions).map("target").uniq()`

src/player.js
endrawes0: Agreed

src/subscription.js
lakinwecker: Agreed.

endrawes0:
Merging.

PR #91, opened 2016-06-28
Switch to lodash

aweidner:
Changed everything over from underscore to lodash and also refactored a bit.

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,20 +8,21 @@
   },
   "dependencies": {
     "async": "2.0.0-rc.3",
+    "async-lock": "0.3.8",
     "botkit": "0.1.1",
     "chai": "3.5.0",
     "codeclimate-test-reporter": "^0.3.3",
+    "eslint-plugin-lodash": "^1.9.2",
     "fast-levenshtein": "1.1.3",
     "google-spreadsheet": "2.0.1",
     "istanbul": "^0.4.3",
+    "lodash": "^4.13.1",
     "moment": "2.13.0",
     "q": "1.4.1",
-    "string-format": "0.5.0",
-    "underscore": "1.8.3",
     "sequelize": "3.23.4",
     "sequelize-cli": "2.4.0",
     "sqlite3": "3.1.4",
-    "async-lock": "0.3.8"
+    "string-format": "0.5.0"
   },
   "devDependencies": {
     "eslint": "^2.13.1",
@@ -29,7 +30,8 @@
   },
   "scripts": {
     "test": "mocha",
-    "send-coverage": "node_modules/.bin/codeclimate-test-reporter < coverage/lcov.info; exit 0"
+    "send-coverage": "node_modules/.bin/codeclimate-test-reporter < coverage/lcov.info; exit 0",
+    "lint": "node_modules/.bin/eslint ."
   },
   "repository": {
     "type": "git",
--- a/src/chesster.js
+++ b/src/chesster.js
@@ -3,7 +3,7 @@ var fs = require('fs');
 var http = require('http');
 var moment = require('moment');
 var Q = require("q");
-var _ = require("underscore");
+var _ = require("lodash");
 
 // Our stuff
 var fuzzy = require('./fuzzy_match.js');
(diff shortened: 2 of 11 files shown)


package.json
```diff
@@ -8,28 +8,30 @@
   },
   "dependencies": {
     "async": "2.0.0-rc.3",
+    "async-lock": "0.3.8",
```
endrawes0: Just realized this was alphabetized... well done.

src/chesster.js
```diff
@@ -422,8 +422,7 @@ chesster.controller.hears([
 	'direct_mention', 
 	'direct_message'
 ], function(bot,message) {
-    bot_exception_handler(bot, message, function(){
-    });
+    bot_exception_handler(bot, message, _.noop)
```
endrawes0: oh i like this. such simply things that I tend to do for myself. I never thought to look for them in a utility.

src/league.js
```diff
@@ -460,11 +460,7 @@ league_attributes = {
     'getCaptains':function() {
         var self = this;
         return Q.fcall(function() {
-            var captains = [];
-            _.each(self._teams, function(team) {
-                captains.push(team.captain);
-            });
-            return captains;
+            return _.map(self._teams, "captain");
```
endrawes0: so this syntax is a little weird to me.

I expect `map` to apply a function to every `team` in `_teams`. 

How should I be reading this?

Clearly it returns an array of captains from `_teams` but it doesnt do this by applying a function - atleast not at this level of abstraction.

src/league.js
```diff
@@ -647,12 +643,13 @@ function League(options) {
 }
 
 function getAllLeagues(config) {
-    var leagues = [];
     var all_league_configs = config['leagues'] || {};
-    _.each(_.keys(all_league_configs), function(key) {
-        leagues.push(getLeague(key, config));
-    });
-    return leagues;
+    return _(all_league_configs)
```
endrawes0: I suppose I should get used to this syntax. I would not have written it this way. I tend to code in a more procedural style - all that C++98...

src/spreadsheets.js
```diff
@@ -407,7 +407,10 @@ function getRows(spreadsheetConfig, options, sheetPredicate, callback) {
                     });
                     var recordRows = [];
                     rows.slice(2).forEach(function(row) {
-                        recordRows.push(_.object(_.zip(headerRow, row)));
+                        var asObject = _.transform(_.zip(headerRow, row), function(result, pair) {
```
endrawes0: I have to learn about lodash. I love the work it does. I would have coded all this manually.

src/subscription.js
```diff
@@ -262,11 +258,7 @@ function getListeners(leagueName, source, event) {
             }
         }).then(function(subscriptions) {
             unlock.resolve();
-            var listeners = [];
-            _.each(subscriptions, function(subscription) {
-                listeners.push(subscription.target);
-            });
-            return _.uniq(listeners);
+            return _(subscriptions).map("target").uniq().value();
```
endrawes0: I love how this reads like a unix pipe.

endrawes0:
Looks great to me. Do as @lakinwecker suggests. I am digging lodash.

src/spreadsheets.js
lakinwecker: This is one area where I think I preferred the underscore version. :(

src/league.js
aweidner: So this syntax is basically [team.captain for team in teams]

src/spreadsheets.js
aweidner: I agree, the underscore version would be much preferred here, unfortunately `_.object` is not defined in lodash and I wasn't able to find a suitable replacement.

src/league.js
endrawes0: That makes sense. Thanks for the explanation.

src/spreadsheets.js
lakinwecker: I found this in lodash :heart: - ```var asObject =_.zipObject(headerRow, row);```

PR #92, opened 2016-06-29
Some refinements to the subscription

lakinwecker:
Adding an event should be far easier now. :)


lakinwecker:
Also fixes some ratings related bugs

src/slack.js
```diff
@@ -318,6 +373,27 @@ function on(options, callback) {
         }));
     });
 }
+//------------------------------------------------------------------------------
+// A helper method to workaround a bug in botkit. 
+//
+// We store a reference to 'bot' in the startRTM method and then use it here
```
endrawes0: Sneaky

src/subscription.js
```diff
@@ -249,24 +244,43 @@ function processRemoveSubscriptionCommand(config, message, id) {
...
 //------------------------------------------------------------------------------
+// Register an event + message handler
+//------------------------------------------------------------------------------
+function register(bot, event_name, cb) {
+    // Ensure this is a known event.
+    events.push(event_name);
+    
+    // Handle the event when it happens
+    emitter.on(event_name, function(league, sources, context) {
+        return getListeners(league.options.name, sources, 'a-game-is-scheduled').then(function(targets) {
```
endrawes0: Should this be `event_name` not a hard coded event?

endrawes0: The hard coded event name it's my only concern here.

endrawes0:
The hard coded event name it's my only concern here.

src/subscription.js
lakinwecker: Yes, it should be!  Good catch.

PR #96, opened 2016-06-30
Link (team) needed to be updated

krcarriere:
Did the same thing as it was prior (links to the Slack file. Not sure if we want this, or want a direct link to the sheet?)


endrawes0:
Link to slack is good

PR #100, opened 2016-07-04
Adding event emission for (1) gamelink and (2) game result parsing. 

endrawes0:
(1) emits a-game-starts and a-game-is-over, depending on whether or not the game has been completed. (2) emits the a-game-is-over event.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,4 +7,5 @@ service_account_key.js
 slack_token.js
 coverage
 test_service_account_key.js
+test_logging_token.js
 db/database.sqlite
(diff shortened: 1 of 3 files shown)


endrawes0:
minor changes are incoming. do not accept this yet

endrawes0:
Considering... should we emit the events when the pairing cache is updated? That way we could enjoy events when someone updates the spreadsheet directly, or leave it as is?

lakinwecker:
Looks good.  

To answer your question. Maybe eventually we should try and detect manual changes to the spreadsheet - but I suspect that's difficult, error prone and hard to implement. So not sure we should do it now.

PR #103, opened 2016-07-05
Fixing the error with the message duplication.

### LINKED ISSUE #102: Incorrect date conversion for notifications
I am UTC-6, but I got this notification today:
```forhavu vs r-mena has been scheduled for 2016-07-06 @ 11:00 UTC, which is Wed @ 13:00 for you.```

Which is (for some reason) converting it it to UTC+2 !? I have tested locally and it's converting properly for me when testing locally. Note that the notification I got for my own game was also incorrect.

``` mooserohde vs lakinwecker has been scheduled for 2016-07-10 @ 20:00 UTC, which is Sun @ 22:00 for you.```

lakinwecker:
Fixes #102.   

Don't share the message between all of the callbacks. :)

### DIFF
--- a/src/subscription.js
+++ b/src/subscription.js
@@ -268,12 +268,17 @@ function register(bot, eventName, cb) {
     // Handle the event when it happens
     emitter.on(eventName, function(league, sources, context) {
         return getListeners(league.options.name, sources, eventName).then(function(targets) {
+            var allDeferreds = [];
             _.each(targets, function(target) {
-                message = cb(target, context);
+                var deferred = Q.defer();
                 bot.startPrivateConversation(target).then(function(convo) {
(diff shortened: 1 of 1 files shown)


endrawes0:
Multi threading is fun

endrawes0:
My understanding of the problem is: different execution contexts used the same instance of message and thus it was a race to see what the value was at the time the conversation was started.

lakinwecker:
In this case, because it's async - there is no race.  The loop would finish completely and the last value of the message would be stored in that context. So all callbacks would use the last value of the message.

So yes, your understanding is correct generally, but because async is a bit more deterministic than threading, there is  a minor subtlety.

endrawes0:
Thanks for the explanation.

PR #113, opened 2016-07-11
Catching exceptions in http and lichess. Logging and rejecting promises.

lakinwecker:


src/lichess.js
```diff
@@ -63,24 +63,53 @@ var makeRequest = (function() {
...
+                    try {
+                        if (result.response.statusCode === 429) {
+                            requestDelay = 60 * SECONDS;
+                            winston.info("[LICHESS] Last request status was a 429 - we will wait 60 seconds before our next request");
+                        } else {
+                            if (mainQueue.length > 0) {
+                                requestDelay = 2 * SECONDS;
+                            } else {
+                                // Back off a bit if we only have background requests.
+                                requestDelay = 4 * SECONDS;
```
endrawes0: Interesting. Any concern that we wont be able to keep up with bg requests with this delay?

lakinwecker: I did some napkin calculations:
30 minutes = 1800 seconds / 4 secs = 450. We can process 450 background requests every 30 minutes.  Why 30 minutes?  Because the only background request we have currently is the background rating update and it happens ever 30ish minutes.  we don't have 450 people in the league.  

Also, they don't publish exact load balancer restrictions, but  lichess did change them temporarily to deal with a DDoS over the weekend, and the 4s is closer to what is allowed during that sort of situation.  

If we run into this, I'm going to be in favour of reducing how often we update ratings, not increasing the rate at which we update them.  But up for discussion, definitely.

endrawes0:
Looks good - just concerned that we may not be able to keep up with bg requests. We update our caches in the mainQueue or bgQueue? If bg, it seems we could take more time getting the data than we do delay between refreshes - how do we avoid the backup? Thanks.

lakinwecker:
The back off to 4s for background requests is not strictly required. It doesn't help us fix the issue this pull request fixes. But it was more of a politeness thing for lichess as its a free service. The BG rating updates are really just for ego sake, I mean honestly who cares if it's your yesterday's rating - do you fluctuate that much daily? 

However, I did sorta throw it in without a ton of thought aside from the napkin calculations I added.  I'm happy to go either way.

endrawes0:
My main concern is whether we can ever catch up or not - if a request is already in the queue, it cannot be added again, correct? if so, I am fine with it. I seem to recall this was the case.

lakinwecker:
We definitely try not to have two requests for a rating in the queue at once:

https://github.com/endrawes0/Chesster/blob/master/src/lichess.js#L163

Lines 163 and 192.

lakinwecker:
Sorry. line 163 and 171

lakinwecker:
Although looking at that, I suspect there is a bug in it.

endrawes0:
Awesome. I thought so. Merging.

endrawes0:
Oh no. Just saw the last message you posted

PR #119, opened 2016-07-27
Fixing scheduling adding logging

lakinwecker:
This adds a test case for one of the recent failing formats and adds some logging. It also changes the spreadsheet call to use a deferred. This would be far more elegant if the spreadsheet API also used a deferred, but I don't have time to do that at the moment.  This should, at least, help us figure out the issue.

At this point, I think it's not a failure to parse the scheduling information that's failing, I suspect that google docs is rate limiting us, or otherwise our google docs requests are failing. Those are a non-promise based API being used within a promise, so I suspect they might be raising an exception or otherwise erroring in a way that isn't being logged or handled.

### DIFF
--- a/src/chesster.js
+++ b/src/chesster.js
@@ -538,17 +538,19 @@ function(bot, message) {
     }
     var schedulingOptions = message.league.options.scheduling;
     if (!schedulingOptions) {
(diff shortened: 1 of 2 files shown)


endrawes0:
Looks fine.

PR #123, opened 2016-07-29
Fixing scheduling adding logging

lakinwecker:
Even more tracing.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -153,7 +153,7 @@ rules:
   max-len: 0
   max-nested-callbacks: 0
   max-params: 0
-  max-statements: [2, 50]
+  max-statements: [2, 60]
   new-cap: 0
   new-parens: 0
   newline-after-var: 0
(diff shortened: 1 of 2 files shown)


.eslintrc
```diff
@@ -153,7 +153,7 @@ rules:
   max-len: 0
   max-nested-callbacks: 0
   max-params: 0
-  max-statements: [2, 50]
+  max-statements: [2, 60]
```
endrawes0: Just 60? I expected to see an order of magnitude ;-)

`max-statements: [2, 500]`

src/chesster.js
```diff
@@ -529,34 +529,47 @@ chesster.on({
     middleware: [slack.withLeague]
 },
 function(bot, message) {
+    var channel = channels.byId[message.channel];
```
endrawes0: Sometimes you have to make a mess worse before it gets better...

endrawes0:
Looks good

.eslintrc
lakinwecker: lol.

PR #131, opened 2016-07-31
Adding more test cases, this proves that it's not the date parsing.

lakinwecker:
This doesn't fix anything, but it proves that it's not the scheduling message format that is causing the issue as all of the ones that I've recorded thus far work without problems in the test.


endrawes0:
That's a great step forward

PR #134, opened 2016-07-31
117 replace flags for player queue with a map player name->promise. 

endrawes0:
if the promise exists then the player is already in the queue.


endrawes0:
fixes 117

lakinwecker:
Looks great.

endrawes0:
updating... then merging

PR #139, opened 2016-08-02
122 some schedule message still not parsing

lakinwecker:


lakinwecker:
Still not what you are looking for. Why did my other changes come with this!?

Issue #122, opened 2016-07-28
Some schedule messages still not parsing


lakinwecker:
Examples: 
````
@joecupojoe @carbon752 7/30 13:30
@joecupojoe vs @carbon752 7/30 13:30
@joecupojoe @carbon752 7/30 @ 13:30 GMT
@joecupojoe vs @carbon752 7/30 @ 13:30 GMT
``

lakinwecker:
MOAR:

```@heidman vs @icendoan 7/30 @ 10:30```

endrawes0:
for us or for slack - idk. but labeling as a bug for now

lakinwecker:
Logs from the most recent error:

Message: `@durchnachtundwind vs @juldiaz280992 Wednesday @ 20:30`

Trace:
```debug: [Scheduling][Message: <@U0J120KAA> vs <@U0UMSDV9A> 08/03 20:30 GMT]: 1. Detecting Scheduling message
debug: [Scheduling][Message: <@U0J120KAA> vs <@U0UMSDV9A> 08/03 20:30 GMT]: 3. Checking Channel
debug: [Scheduling][Message: <@U0J120KAA> vs <@U0UMSDV9A> 08/03 20:30 GMT]: 5. Checking Scheduling Options
debug: [Scheduling][Message: <@U0J120KAA> vs <@U0UMSDV9A> 08/03 20:30 GMT]: 6. Checking Spreadsheet Options
debug: [Scheduling][Message: <@U0J120KAA> vs <@U0UMSDV9A> 08/03 20:30 GMT]: 7. Checking if channel is in schedulingOptions
debug: [RESULTS][Message: <@U0J120KAA> vs <@U0UMSDV9A> 08/03 20:30 GMT]: 1. Checking for League
debug: [RESULTS][Message: <@U0J120KAA> vs <@U0UMSDV9A> 08/03 20:30 GMT]: 2. Checking for channel
debug: [RESULTS][Message: <@U0J120KAA> vs <@U0UMSDV9A> 08/03 20:30 GMT]: 3. Checking for spreadsheetOptions
debug: [RESULTS][Message: <@U0J120KAA> vs <@U0UMSDV9A> 08/03 20:30 GMT]: 4. Checking for resultsOptions
debug: [GAMELINK][Message: <@U0J120KAA> vs <@U0UMSDV9A> 08/03 20:30 GMT]: 1. Detecting gamelink
[...]

PR #141, opened 2016-08-03
in single-purpose channels, ONLY use channel name for league detection.

lakinwecker:
This should fix the issue we've been seeing. I don't have all of the messages that where failing, but the ones that were failing were due to chesster thinking that the message text was identifying the lone wolf league as a target.  In the single-purpose channels, we don't want that. posts in #team-scheduling should always be considered a 45+45 message, regardless of content. 

This achieves that.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -153,7 +153,7 @@ rules:
   max-len: 0
   max-nested-callbacks: 0
   max-params: 0
-  max-statements: [2, 600]
+  max-statements: [2, 50]
   new-cap: 0
   new-parens: 0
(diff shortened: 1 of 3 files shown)


lakinwecker:
I believe this closes #122

endrawes0:
Makes sense to me. Will review when I'm fully awake

endrawes0:
That middleware is proving to be as beautifully flexible as I had originally hoped.

endrawes0:
Looks great. Thanks. 

I'm surprised we missed the pattern of games scheduled on the half hour.

PR #155, opened 2016-09-18
availability and jshint

endrawes0:
Added a couple functions that I want to apply across the board.

stripComponents and formatIsValid

These take care of a common idiom that I saw in command parsing of creating variables by name and comparing them to their expected values.

See how they are used and how they simplify some of the code and reduce hardcoding, in a way.

If approved, I will apply across the board.

There are minor changes made across to code base to clean things up after adding jshint.

I also removed some extraneous error checking that should be handled in the middleware.

### DIFF
--- a/src/chesster.js
+++ b/src/chesster.js
@@ -154,7 +154,7 @@ function(bot, message){
         ];
     }
 
-    commands.tokenize(message.text, commandDescription).then(function(parameters){
+    return commands.tokenize(message.text, commandDescription).then(function(parameters){
         var available = parameters["available"] ? true : false;
         var roundNumber = parameters["roundNumber"];
         var speaker = slack.getSlackUserFromNameOrID(message.user);
@@ -182,7 +182,7 @@ function(bot, message){
             }
         }
 
-        heltour.setAvailability(heltourOptions, playerName, available, roundNumber).then(function (){
+        return heltour.setAvailability(heltourOptions, playerName, available, roundNumber).then(function (){
             formatReplyUpdatedAvailability(bot, message, playerName, available, roundNumber);
         }).catch(function(error){
             replyFailedToUpdate(bot, message, "availability", error); 
@@ -263,7 +263,7 @@ function(bot, message){
         return;
     }
     
-    commands.tokenize(message.text, [
+    return commands.tokenize(message.text, [
         "assign", 
         "{text:player}", 
         "to", 
@@ -296,7 +296,7 @@ function(bot, message){
         if(!player){
             replyFailedToUpdate(bot, message, "alternate assignment", "unknown player");
         }
-        heltour.assignAlternate(heltourOptions, 
+        return heltour.assignAlternate(heltourOptions, 
             roundNumber, 
             team.number, 
             boardNumber, 
@@ -404,7 +404,7 @@ function(bot, message){
         replyUnrecognizedTeam(bot, message, teamName);
         return;
     }
-    heltour.assignAlternate(heltourOptions, roundNumber, team.number, boardNumber, player).then(function(){
+    return heltour.assignAlternate(heltourOptions, roundNumber, team.number, boardNumber, player).then(function(){
         bot.reply(message, player + " has been assigned to board " + boardNumber + " for " + teamName + " during round " + roundNumber);
     }).catch(function(error){
         replyFailedToUpdate(bot, message, "alternate assignment", error);
@@ -1058,7 +1058,7 @@ function(bot, message) {
             return;
         }
         
-        heltour.findPairing(
+        return heltour.findPairing(
             heltourOptions,
             result.white.name,
             result.black.name,
@@ -1293,7 +1293,7 @@ function processGameDetails(bot, message, details, options, heltourOptions){
     //gamelinks only come from played games, so ignoring forfeit result types
 
     //update the spreadsheet with results from gamelink
-    heltour.updatePairing(
+    return heltour.updatePairing(
         heltourOptions,
         result
     ).then(function(updatePairingResult) {
--- a/src/chesster.js
+++ b/src/chesster.js
@@ -118,6 +118,169 @@ function leagueDMResponse(patterns, responseName) {
 leagueResponse(['captain guidelines'], 'formatCaptainGuidelinesResponse');
 leagueDMResponse(['captains', 'captain list'], 'formatCaptainsResponse');
 
+/* 
+ * strip components
+ * Splits the text. Adds each word to an object using the keys provided. 
+ *
+ * Each value, component, is made to be lower case.
+ *
+ * If the number of components is fewer than the number of words in the text,
+ * the remaining words will be joined as the final argument.
+ */
+function stripComponents(text, componentKeys){
+    var numComponents = componentKeys.length;
+    var toStrip = _.toLower(text);
+
+    var tokens = _.split(toStrip, " ");
+    var args = _.slice(tokens, 0, componentKeys.length-1);
+    var lastArg = _.slice(tokens, componentKeys.length-1).join(" ");
+
+    if( args.length > numComponents ) 
+        throw new Error("Too few arguments to strip");
+
+    var components = {}; //add each argument to the object, using the keys given
+    _.each(args, function(argument, index){
+        components[componentKeys[index]] = tokens[index];
+    });
+    components[componentKeys[numComponents-1]] = lastArg;
+    
+    return components;
+}
+
+/* availability */
+/* [player <player-name> is] {available, unavailable} for round <round-number> in <league> */
+chesster.hears({
+    middleware: [ 
+        slack.withLeague
+    ],
+    patterns: [
+        'available',
+        'unavailable'
+    ],
+    messageTypes: [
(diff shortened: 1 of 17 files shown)


src/chesster.js
```diff
@@ -118,6 +118,169 @@ function leagueDMResponse(patterns, responseName) {
...
+/* 
+ * strip components
+ * Splits the text. Adds each word to an object using the keys provided. 
+ *
+ * Each value, component, is made to be lower case.
+ *
+ * If the number of components is fewer than the number of words in the text,
+ * the remaining words will be joined as the final argument.
+ */
+function stripComponents(text, componentKeys){
```
lakinwecker: I'm not a big fan of the name "stripComponents".  I think I would prefer the name 'tokenize'.

src/chesster.js
```diff
@@ -118,6 +118,169 @@ function leagueDMResponse(patterns, responseName) {
...
+    if(!heltourOptions){
+        winston.error("{} league doesn't have heltour options!?".format(message.league.options.name));
+        return;
+    }
+
+    // I hope I am not over-using the ternary operator here. 
+    // I would like to apply stripComponents to the other functions.
+    // but only if we agree this cleans things up.
+    var components = 
+        (! message.text.includes("player")) ?
```
lakinwecker: This will not work properly if a team name includes "player" or a player name includes player: https://en.lichess.org/@/player

src/chesster.js
```diff
@@ -118,6 +118,169 @@ function leagueDMResponse(patterns, responseName) {
...
+    }
+
+    // I hope I am not over-using the ternary operator here. 
+    // I would like to apply stripComponents to the other functions.
+    // but only if we agree this cleans things up.
+    var components = 
+        (! message.text.includes("player")) ? 
+        stripComponents(
+            message.text, 
+            [ 'available', 'for', 'round', 'roundNumber', 'in','league' ]
```
lakinwecker: This is much better than what we had before, but I think we could go one step further and do something like: 
```
 [ 'available', 'for', 'round', '{roundNumber}', 'in','{league}' ]
```
or
```
 [ 'player', '{playerName}', 'is' 'available', 'for', 'round', '{roundNumber}', 'in','{league}' ]
```

In this case, the { denotes that it's a variable and will take whatever token appears there, while the absence of "{" will indicate that it _must_ be that value and it can throw an error if it's not.

We could take it further and use '|' for two options that can appear at the same spot.

Also, the function could take multiple lists instead of just one and return the first one that it finds that matches, which would remove the need for the ternary operator here. 

Perhaps I'm making the function too complex, but I think that at least the initial idea (denoting which are variable and which are constants) will help

src/chesster.js
```diff
@@ -118,6 +118,169 @@ function leagueDMResponse(patterns, responseName) {
...
+        + "*@{playerName} will{not}* be available for *round{roundNumber}*."
+            .format({ 
+                "playerName": playerName,
+                "not": (available ? "" : " not"),
+                "roundNumber": " " + roundNumber
+            }
+        )
+    );
+}
+//I am thinking this is easier.
```
lakinwecker: I find them both equally easy to read, so go with whichever you prefer.

src/chesster.js
```diff
@@ -118,6 +118,169 @@ function leagueDMResponse(patterns, responseName) {
...
+function formatReplyUpdatedAailabilty(bot, message, playerName, available, foundNumber){
```
lakinwecker: If you go with this one, the last word is missing a 'v'.

src/chesster.js
lakinwecker: If you go with this style of parsing, I still think you should return the value of the constants that you detected so that you can do things like:
```
var components = tokenize(message.text, ['unavailable|available', 'for', 'round', '{roundNumber}', 'in', '{league}'])
if (components['available']) {
...
} else if (components['unavailable']) {
...
}

File: src/chesster.js [hunk 6]
```diff
@@ -894,13 +1032,6 @@ chesster.on({
     middleware: [slack.withLeagueByChannelName]
 },
 function(bot, message) {
-    var channel = channels.byId[message.channel];
```
lakinwecker: I agree. These should be middleware.

src/commands.js
```diff
@@ -0,0 +1,183 @@
...
+    //create an object splitting the description string into type and name
+    var description = 
+        _.zipObject(['type', 'name'], _(descriptionString).split(/[\{\:\}]/).filter().value());
+    //if either are not specified, the description string is bad.
+    if(_.isNil(description.type) || _.isNil(description.name)) 
+        throw new InvalidTokenDescriptionError(descriptionString);
+    //for each type defined, test it to see if it matches
+    //   given a match, parse the token and store
+    _.forIn(VARIABLE_TYPES, function (type){
+       if(description.type.match(type.match_expression)){
```
lakinwecker: Is there a reason to use a regex match here instead of a lookup?

endrawes0: i dont think so. just regex happy. you are correct a simple look up would be much more straight forward.

src/commands.js
```diff
@@ -0,0 +1,183 @@
...
+
+function InvalidTokenDescriptionError(description){
+    this.description = description;
+    this.name = 'InvalidTokenDescriptionError';
+    this.stack = (new Error()).stack;
+}
+
+/*
+ * separates the text into numTokens-1, where the remaining words make up the final token
+ * if the number of tokens is greater than the number of " " separted words, the array of
```
endrawes0: k

src/commands.js
```diff
@@ -0,0 +1,183 @@
...
+                parameterizeConstantToken(token, description, parameters); 
+            }else{
+                throw new InvalidTokenDescriptionError(description);
+            }
+        });
+        return parameters;
+    });
+}
+
+function TooFewTokensError(tokens, descriptions){
```
endrawes0: Yeah - I did that for convenience at first but should have relocated them after. I will move them.

src/commands.js
```diff
@@ -0,0 +1,183 @@
...
+ *           InvalidConstantError
+ * variable: {type: variableName} - variable values will be varified to be the 
+ *           type specified. If the type does not match, this function will throw
+ *           InvalidTypeValueError
+ * choice:   {option1|option2|option3} - choices must be one of the prespecified  
+ *           values. The option will be stored in the key and value of the 
+ *           returned token object. If it does not match one of the specified 
+ *           options, this function will throw InvalidChoiceError
+ * 
+ * types:    int, text - type must match one of these. If it does not match, thi
```
endrawes0: k

src/commands.js
```diff
@@ -0,0 +1,183 @@
...
+    this.name = 'InvalidTypeValueError';
+    this.stack = (new Error()).stack;
+}
+
+/* 
+ * This command tokenizer takes the following token descriptions
+ * constant: constantValue - constant values will be varified to be the exact 
+ *           value specified. If it does not match, this function will throw 
+ *           InvalidConstantError
+ * variable: {type: variableName} - variable values will be varified to be the
```
endrawes0: k

src/chesster.js
```diff
@@ -118,6 +108,137 @@ function leagueDMResponse(patterns, responseName) {
...
+            "in",
+            "{text:league}"
+        ];
+    }
+
+    commands.tokenize(message.text, commandDescription).then(function(parameters){
+        var available = parameters["available"] ? true : false;
+        var roundNumber = parameters["roundNumber"];
+        var speaker = slack.getSlackUserFromNameOrID(message.user);
+        var playerName = speaker.name;
```
endrawes0: not a bad idea - gets my mind thinking about other possibilities. thanks

src/chesster.js
```diff
@@ -145,63 +263,73 @@ function(bot, message){
...
-        board       = args[3],
-        boardNumber = args[4], 
-        during      = args[5], 
-        round       = args[6],
-        roundNumber = args[7],
-        on          = args[8];
-
-    var speaker = slack.getSlackUserFromNameOrID(message.user);
-    var speakerTeam = message.league.getTeamByPlayerName(speaker.name);
+    commands.tokenize(message.text, [
```
endrawes0: what are we referring to here? I am not sure what you want returned.
thanks.

src/chesster.js
```diff
@@ -118,6 +108,137 @@ function leagueDMResponse(patterns, responseName) {
...
+    var heltourOptions = message.league.options.heltour;
+    if(!heltourOptions){
+        winston.error("{} league doesn't have heltour options.".format(message.league.options.name));
+        return;
+    }
+
+    var commandDescription;
+    if(_.isEqual(_(message.text).split(' ').head(), "player")){
+        commandDescription = [
+            "player",
```
endrawes0: It has been a fun exercise so I don't regret our choices.

src/commands.js
endrawes0: idk why I kept mispelling that... I did it twice...

src/chesster.js
lakinwecker: The fact that we can add a playerName type to our system makes me like our system.

PR #170, opened 2016-09-29
Bug in captain alt assignment

endrawes0:
I thought I tested the captains stuff but bc I was using the wrong isModerator function (used slack not league), I never actually hit that code.


tnan123:
does this bug still exist? or is this a different issue?

Some captains still get the message
chessterBOT [03:28]  
Only the team's captain or a league moderator can do that.

PR #175, opened 2016-10-16
Commandsjs unit tests

endrawes0:
Finally finished these tests.

### DIFF
--- a/test/test_commands.js
+++ b/test/test_commands.js
@@ -6,7 +6,7 @@ describe('commands', function(){
         function assertIfCalled(message){
             return function(){ assert(false, message); };
         }
-        it("should tokenize a constant string as an object containing parameters for each constant, of the same name and with the constant as the value", function(){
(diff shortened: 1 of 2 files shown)


test/test_commands.js
```diff
@@ -0,0 +1,111 @@
...
+                    'constant': 'constant',
+                    'intVar': 4,
+                    'textVar': 'text',
+                    'option1': 'option1',
+                });
+            });
+        });
+        it("throw an InvalidConstantError if the token does not exactly match the specified constant", 
+            function(){
+                return commands.tokenize('this', ['that']).then(assertIfCalled('tokenize did not throw'))
```
lakinwecker: I would prefer a better error message here. "Did not throw when a constant doesn't not match the input string" or some such.

lakinwecker: This goes for all of the "did not throw" below

endrawes0: Yea - I was being lazy - 

Thanks for keeping me honest.

Issue #176, opened 2016-10-17
Require a gamelink when posting a result


cyanfish:
Users shouldn't be able to post results without posting a gamelink first. Moderators should be able to post results in any case though.

PR #177, opened 2016-10-19
Moderator Fixes / Requests

endrawes0:
Comparisons against moderators are now lower case. Added handling for results without gamelinks. Moderators can still post results without a link.

### DIFF
--- a/src/chesster.js
+++ b/src/chesster.js
@@ -1064,7 +1064,7 @@ function(bot, message) {
         if(
             !_.isEqual(result.white.id, message.user) &&
             !_.isEqual(result.black.id, message.user) &&
-            !message.player.isModerator(message)
+            !message.player.isModerator()
         ){
             replyPermissionFailure(bot, message);
             return;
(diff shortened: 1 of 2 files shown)


endrawes0:
fixes #176

src/chesster.js
```diff
@@ -1095,37 +1095,41 @@ function(bot, message) {
                     result);
             }else if(!_.isNil(pairing)){
                 //update the pairing with the result bc there was no link found
```
lakinwecker: This comment deserves to be updated

src/chesster.js
```diff
@@ -1064,7 +1064,7 @@ function(bot, message) {
         if(
             !_.isEqual(result.white.id, message.user) &&
             !_.isEqual(result.black.id, message.user) &&
-            !message.player.isModerator()
+            !message.player.isModerator(message)
```
lakinwecker: I don't understand why message was required here. But I also haven't looked at where this is defined.

src/chesster.js
endrawes0: good call.

src/chesster.js
endrawes0: it compares it to the league associated with the message.

a little odd - just a helper on the player. Im not sure why or when we added it. but there it is.

PR #185, opened 2016-11-04
This should allow the version that is documented to work.

lakinwecker:
Along with a few other permutations that seem relatively natural.


lakinwecker:
Yeah, it was great!

PR #186, opened 2016-11-04
Lowercase the player name before we store the rating, so the lookup works

lakinwecker:


endrawes0:
Getting server error when I attempt to merge. 

Will try again later.

lakinwecker:
It's because this branch is now out of date with the base branch. I'll rebase

PR #188, opened 2016-11-05
Fiddling with timezones

lakinwecker:


lakinwecker:
Yup. Provided you can get your hands on the timezone identifiers, and an olsen database, you can solve most timezone issues reasonably well. There are still edge cases that won't work, but most will and this is probably good enough for chesster

lakinwecker:
Can I deploy this?

config/testconfig.js
```diff
@@ -33,12 +33,13 @@ config["leagues"]["lonewolf"]["heltour"]["baseEndpoint"] = "https://staging.lich
 config["leagues"]["45+45"]["scheduling"]["channel"] = "unstable_bot";
 config["leagues"]["45+45"]["results"]["channel"] = "unstable_bot";
 config["leagues"]["45+45"]["gamelinks"]["channel"] = "unstable_bot";
-config["leagues"]["45+45"]["alternate"]["channel_id"] = "G2G6URBT5";
+config["leagues"]["45+45"]["alternate"]["channel_id"] = "C0VCCPMJ8";
 config["leagues"]["lonewolf"]["scheduling"]["channel"] = "unstable_bot-lonewolf";
 config["leagues"]["lonewolf"]["results"]["channel"] = "unstable_bot-lonewolf";
 config["leagues"]["lonewolf"]["gamelinks"]["channel"] = "unstable_bot-lonewolf";
 
 config["channel_map"]["G2G6URBT5"] = "45+45";
```
kuc: Shouldn't this line be deleted after changes made to line 36?
BTW I don't even know what changes in this file have with purpose of this PR. You made them intentionally?

src/chesster.js
```diff
@@ -850,8 +850,8 @@ function schedulingReplyAmbiguous(bot, message){
 
 // Game has been scheduled.
 function schedulingReplyScheduled(bot, message, results, white, black) {
-    var whiteDate = results.date.clone().utcOffset(white.tz_offset/60);
-    var blackDate = results.date.clone().utcOffset(black.tz_offset/60);
+    var whiteDate = results.date.clone().tz(white.tz);
+    var blackDate = results.date.clone().tz(black.tz);
```
kuc: Nice, but you forgot about `function localTime(datetime)` in file `slack.js`

kuc: Also, how about removing notice from commit 37ab60cef, since it's no longer (hopefully) needed?

config/testconfig.js
lakinwecker: Yes, I will remove. I was changing testing channels. I will review the file and fix it.

kuc: So I suggest reverting all changes to this file (and eventually putting them to separate PR).

src/chesster.js
lakinwecker: Agreed. I will update the localTime Function.

I will remove the warning. I'm not 100% convinced that it's not useful, but this changeset should make chesster as accurate as he ever was in terms of time zone conversions.

kuc: We could just change content of the warning, instead of removing it entirely. _Something something, pay attention to incoming DST changes, blah blah_

lakinwecker: I've removed the message. This is as accurate as we can make it. The only reason we will be wrong is if the person gets their slack timezone wrong. I'm ok with that situation.

lakinwecker:
I have added the same code to the other localTime function and removed the unnecessary changes to the config. Please re-review.

lakinwecker:
In this case, the localTime function used in slack does not have control over the type passed in, so I also changed the require to use moment-timezone across the board. This should ensure that we have the .tz method available in all cases.

This is ready for review again.

kuc:
Sorry for delay, I will review this tomorrow

package.json
```diff
@@ -19,6 +19,7 @@
     "jshint": "^2.9.3",
     "lodash": "^4.13.1",
     "moment": "2.13.0",
```
kuc: Is that library still used by anything, after migration to `moment-timezone`?

lakinwecker:
moment removed from package.json.

PR #190, opened 2016-11-24
Updates to setup process

arwagner:
- Sample token files
- Update README
- Get rid of references to files that are not needed
- etc.


config/testconfig.js
```diff
@@ -18,12 +17,10 @@ config["welcome"]["channel"] = "unstable_bot";
 config["leagues"]["45+45"]["spreadsheet"]["key"] = "1BeRN76zaB_uCCrCra2yTEEw_r6C5_b8P59aN_BrJsyA";
 config["leagues"]["45+45"]["spreadsheet"]["serviceAccountAuth"] = {
```
lakinwecker: Can we completely get rid of serviceAccountAuth? I don't think we use the spreadsheets at all anymore, right?

config/config.js
```diff
@@ -55,8 +54,7 @@ var config = {
             "spreadsheet": {
                 "key": "1cTBOQNrbUw9-RWmrC9b6Yyi3i2MtIvUWuXNxzQmGfSs",
                 "serviceAccountAuth": {
```
lakinwecker: Sam with here.

config/config.js
```diff
@@ -55,8 +54,7 @@ var config = {
             "spreadsheet": {
```
endrawes0: I wont ask you to do this, but we should be able to drop the entire spreadsheet section of the config - as soon as we remove the dependencies on src/spreadsheets.js

config/testconfig.js
endrawes0: confirmed - its unneeded 👍

config/config.js
endrawes0: confirmed - its unneeded

endrawes0:
Thanks for the contribution - will help significantly with onboarding new devs.

PR #192, opened 2016-11-24
make 'notify' a synonym of 'summon'

arwagner:


src/chesster.js
```diff
@@ -547,6 +547,7 @@ function(bot,message) {
 /* mods */
 
 leagueResponse(['summon mods'], 'formatSummonModsResponse');
+leagueResponse(['notify mods'], 'formatSummonModsResponse');
```
endrawes0: well done. very elegant solution ;)

lakinwecker: I believe the reason why that first parameter is a list is that you can pass in multiple regexes there.

lakinwecker: Meaning you can simplify that to: `leagueResponse(['summon  mods', 'notify mods', 'holla at mods'], 'formatSummonModsResponse');`

PR #200, opened 2016-12-07
Backoff in the case of errors

lakinwecker:
Try twice within 10s and we get errors in that context, back off and wait for the next refresh.

### DIFF
--- a/config/config.js
+++ b/config/config.js
@@ -30,7 +30,7 @@ var config = {
         token: logging_token,
         username: "chesster",
         level: "debug",
-        handleExceptions: false
+        handleExceptions: true
     },
     "links": {
         "source": "https://github.com/endrawes0/Chesster"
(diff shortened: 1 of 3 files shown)


lakinwecker:
Thanks - I'll work on these in a few minutes

lakinwecker:
Ok, issues addressed. There is a self.req = null in the .on ('end' version, but I think that one is ok, because .abort() is not needed in that case. 

Although I must admit my knowledge of how these events play out are not as good as they should be.

lakinwecker:
As it turns out this is not the right pattern either. When `self.req.abort()` is called, the `.on('end')` version is also called. Resulting in more requests than necessary, and specifically triggering the error handling faster than normal.

I'm going to call `self.req.abort(); self.req = null` in the `catch` clause, which should end the request and the `.on('end')` handler will start the watcher again.

PR #202, opened 2016-12-13
Break up main chesster file

arwagner:
This shouldn't be merged - it's just to suggest the next logical step in the command re-org, for discussion

### DIFF
--- a/src/chesster.js
+++ b/src/chesster.js
@@ -16,127 +16,16 @@ var chesster = new slack.Bot({
     config_file: config_file
 });
 
-const games = require('./commands/games.js')(chesster);
-const availability = require("./commands/availability.js");
-const nomination = require("./commands/nomination.js");
-const scheduling = require("./commands/scheduling.js");
-const leagueInfo = require("./commands/leagueInfo.js");
-const onboarding = require("./commands/onboarding.js");
-const playerInfo = require("./commands/playerInfo.js");
-const subscription = require('./commands/subscription.js');
-
-// A helper for a very common pattern
-function directRequiresLeague(patterns, callback) {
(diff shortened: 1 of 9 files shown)


src/commands/subscription.js
```diff
@@ -1,428 +1,452 @@
...
-module.exports.register = register;
-
-module.exports.tellMeWhenHandler = tellMeWhenHandler;
-module.exports.helpHandler = helpHandler;
-module.exports.listHandler = listHandler;
-module.exports.removeHandler = removeHandler;
-module.exports.formatAGameIsScheduled = formatAGameIsScheduled;
-module.exports.formatAGameStarts = formatAGameStarts;
-module.exports.formatAGameIsOver = formatAGameIsOver;
+module.exports = function(chesster, slack){
```
lakinwecker: You can just import slack here.  You don't need to pass it in.

src/chesster.js
```diff
@@ -24,118 +16,16 @@ var chesster = new slack.Bot({
...
-chesster.hears(
-    {
-        patterns: [
-            slack.appendPlayerRegex("pairing", true)
-        ],
-        messageTypes: ['direct_mention', 'direct_message']
-    },
-    playerInfo.playerPairings(chesster.config)
-);
+const games = require('./commands/games.js')(chesster, slack);
```
lakinwecker: Same goes for all of these, slack is just another module. I'd rather you just imported it in the commands modules directly.

lakinwecker:
My personal preference would be to export a register function in addition to the other functions, that way we can properly requires the functions we need to test, and the registration is explicit.

PR #204, opened 2016-12-17
add a basic addChallenge command for the ladder

arwagner:


src/commands/ladder.js
```diff
@@ -0,0 +1,17 @@
...
+const slack = require("../slack");
+
+function addChallenge(config, logic){
+    return function(bot, message){
+        if (_.isEqual(message.channel, config["mod-channel"])){
+	    var words = message.text.split(" ");
+	    if (words.length != 5){
+                bot.reply(message, "usage: @chesster add challenge <challenger> <defender> <white>");
+            } else {
+		logic.addChallenge(words[2], words[3], words[4]);
```
lakinwecker: Chesster uses promises. If you return a promise from here then any errors in the callback chain will be handled properly.  At the moment, if there is an exception in the spreadsheet callbacks, that they don't handle, it won't be logged and chesster may crash (I haven't figured out if this is consistent behavior and he always crashes, or only some times of errors/exceptions cause crashes).

Again, this won't stop me from merging/deploying as this command is only for mod usage, so I don't expect you guys to cause this to happen often. :)

lakinwecker:
Please fix the lint issues identified in the CI failure.  You can run the linter locally with: 

`npm run lint`

And run jshint with: 
`npm run hint`

Once the build passes, I will happily merge and deploy this for you. Good work. I'm excited to see this go live and start helping you guys out.

PR #205, opened 2016-12-18
 Initial version of the forwardingMessage command. 

lakinwecker:
This required a bit more refactoring than I originally anticipated. One
of the things I had to do is refactor the slack module to allow it to
be used more than once, which resulted in my having to touch a lot of
code.  Also, I've had to change it so that we are passing the instance
of our bot object into the .on / .hears callback - so I've had to make
that instance work like the botkit.bot.

### DIFF
--- a/src/league.js
+++ b/src/league.js
@@ -663,6 +663,7 @@ var league_attributes = {
 function League(bot, options) {
     this.options = {};
     this.emitter = new ChessLeagueEmitter();
+    this.bot = bot;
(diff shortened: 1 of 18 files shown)


src/league.js
```diff
@@ -661,26 +660,26 @@ var league_attributes = {
     }
 };
 
-function League(options) {
+function League(bot, options) {
```
cyanfish: Missing an assignment to this.bot?

lakinwecker: Fixed.

PR #210, opened 2016-12-21
fix 2 small bugs in the ladder functionality

arwagner:


src/spreadsheet.js
```diff
@@ -27,7 +27,7 @@ module.exports = function(key, creds){
             }, function(err, cells){
                 for (var i = 0; i < data.length; i++){
     	        cells[i].value = data[i];
-    		cells[i].save(_.noop());
+    		cells[i].save(_.noop);
```
lakinwecker: Are you sure that's a bug? That's a method which returns undefined. Calling it here is the same as passing in undefined, which is different from what you originally had (which is an empty function).

PR #234, opened 2017-01-24
Testing @seb32 ping

Sebastien32:
hi endrawes

### DIFF
--- a/src/watcher.js
+++ b/src/watcher.js
@@ -143,7 +143,7 @@ function Watcher(bot, league) {
             self.bot.say({
                 text: "<@" + white + ">,  <@" + black + ">:"
                     + " Your game is *not valid* because "
-                    + "*" + result.reason + "*",
+                    + "*" + result.reason + "*" + "@seb32",
                 channel: self.league.options.gamelinks.channel_id
             });
             self.bot.say({


endrawes0:
Please add a summary and description

endrawes0:
Also, this breaks our lint rules. I'll explain inline.

src/watcher.js
```diff
@@ -143,7 +143,7 @@ function Watcher(bot, league) {
             self.bot.say({
                 text: "<@" + white + ">,  <@" + black + ">:"
                     + " Your game is *not valid* because "
-                    + "*" + result.reason + "*",
+                    + "*" + result.reason + "*" + "@seb32",
```
endrawes0: The linter is complaining about this line.

It was you to remove the unnecessary concatenation.
 
`"...*"+"@seb32"`

Replace with a single string constant.

endrawes0: Also, this won't work unless you have notifications on for strings matching your name.

You need to use the user ID and Mark up to identify it as a special token.

Sebastien32: ok, will check how the other ones are done

endrawes0: 👍

PR #237, opened 2017-02-26
Add a "refresh pairings" command to msg-forward

cyanfish:


src/commands/messageForwarding.js
```diff
@@ -107,4 +108,41 @@ function forwardMessage(chesster, adminSlack) {
...
+                return;
+            }
+
+            if (type === "pairings") {
+                _league.refreshCurrentRoundSchedules().then(function(pairings) {
+                    winston.info("Found " + pairings.length + " pairings for " + _league.options.name + " (manual refresh)");
+                }).catch(function(error) {
+                    winston.error("{}: Error refreshing pairings: {}".format(_league.options.name, JSON.stringify(error)));
+                });
+            }
```
lakinwecker: It probably doesn't really matter - but we could log what the other `type` is here in case something gets messed up.

PR #240, opened 2017-04-01
Fixing the game starts and game ends commands

lakinwecker:


src/commands/subscription.js
```diff
@@ -302,6 +302,8 @@ function register(bot, eventName, cb) {
                     var message = cb(bot, target, _.clone(context));
                     convo.say(message);
                     deferred.resolve();
+                }).catch(function(error) {
+                    console.error(error);
```
cyanfish: Should be winston.error

lakinwecker: Fixed.

PR #241, opened 2017-04-02
Allow my-team-channel as a target and my-team as a source

lakinwecker:


src/commands/subscription.js
```diff
@@ -32,16 +32,19 @@ function formatHelpResponse(bot, config) {
...
+            "3. `subscription remove <id>` to list remove a specific subscription\n" +
+            "\n--------------------------------------------------------------------\n" +
+            "For command 1, these are the options: \n" + 
+            "1. `<listener>` is either `me` to subscribe yourself or " +
+            " `my-team-channel` to subscribe your team channel (_Note: you must be " +
+            " be team captain to use `my-team-channel`_).\n" +
+            "2. `<event>` is one of:`" + events.join("` or `") + "`\n" +
+            "3. `<target-user-or-team>` is the target username or teamname that you are " +
+            "subscribing to (_Note: You can use `my-team` to target your team_).\n" +
+            "4. `<league>` is one of: `" + leagueNames.join("` or `") + "`";
```
cyanfish: I'd switch the order of 3 and 4 to match the order in the command

src/commands/subscription.js
```diff
@@ -298,13 +325,22 @@ function register(bot, eventName, cb) {
...
-                bot.startPrivateConversation(target).then(function(convo) {
-                    var message = cb(bot, target, _.clone(context));
-                    convo.say(message);
-                    deferred.resolve();
-                }).catch(function(error) {
-                    winston.error(JSON.stringify(error));
-                });
+                var message = cb(bot, target, _.clone(context));
+                if (_.startsWith(target, 'channel_id:')) {
+                    var channelID = _.toUpper(target.substr(11));
```
cyanfish: I don't think toUpper is needed here

src/commands/subscription.js
```diff
@@ -236,10 +259,14 @@ function processSubscriptionListCommand(bot, config, message) {
...
                     ['id', 'ASC']
                 ]
-			}).then(function(subscriptions) {
+            }).then(function(subscriptions) {
                 var response = "";
                 _.each(subscriptions, function(subscription) {
-                    response += "\nID {id} -> tell {target} when {event} for {source} in {league}".format(subscription.get());
+                    var context = subscription.get();
+                    if (_.startsWith(context['target'], 'channel_id:')) {
+                      context['target'] = 'your team channel';
```
cyanfish: This could display the channel name (e.g. "#team-1-s7") to be a bit more general

src/commands/subscription.js
lakinwecker: done.

src/commands/subscription.js
lakinwecker: I made a quick pass at trying to get that, but I'm not sure chesster always has access to that information. Does he get the channel names for all channels he is in?

src/commands/subscription.js
lakinwecker: It is.

src/commands/subscription.js
lakinwecker: Yeah, I just tried again. The secret dev channel that I'm using is not in Chesster's list. Maybe we could get it from somewhere - but at the moment, we can't turn that id into a name.

cyanfish: Right, it'd actually a group, so we'd need to add a groups property and a call to `groups.list`... May not be worth it for this little thing

src/commands/subscription.js
cyanfish: Got it, I missed the toLowerCase when it was saved to the DB

PR #283, opened 2018-07-04
Switch to postgresql, drop lock

lakinwecker:


cyanfish:
Might want to remove async-lock from package.json, but the code looks good.

PR #303, opened 2020-03-08
Partial port to typescript and minimizing watcher requests

lakinwecker:
As per a recent change to Lila, we can only have two watcher requests open at a time with a maximum of 300 users per request. 

This change refactors the watcher to comply with this request and starts down the road of porting to typescript.


lakinwecker:
Closing in favour of the complete port

PR #304, opened 2020-04-11
Complete port to typescript 

lakinwecker:
This is ready for a review.  I have a few minor things to test over the next week / after it's deployed.

It's a complete port to typescript as well as removal of botkit in exchange for directly using the official Slack RTM and web API clients.   I was able to get rid of all anys except in a few cases where it makes sense (arbitrary JSON responses).  I was also able to get type safe JSON parsing working which should inform us of errors when things change etc.

I've also updated the unit tests to cover more things, and removed old lint/hinting tools and added in a new tslint tool.  I've also updated it with a .editorconfig and .prettier configs so that we can have consistent automated formatting.  

Finally, I have gone through and tested each command at least once. I can't guarantee that I've tested every last code path, but I've tested most.


src/commands.ts
```diff
@@ -0,0 +1,222 @@
...
+ * get the type of the variable from the description
+ * call the specific token verification routine for the type specified
+ * get the key from the description for parameterization
+ * add the variable to the parameter object using the key from the description
+ * if the name or type cannot be determined from the dscription, throw InvalidTokenDescription
+ * if the type specified is not known, throw InvalidVariableTypeError
+ */
+function parameterizeVariableToken(
+    token: string,
+    dscription: string,
```
cyanfish, line 105: Variable name - maybe call this "description" and deconstruct "type" and "name" directly?

src/commands.ts
```diff
@@ -0,0 +1,222 @@
...
+
+export class TextVariable {
+    typeName = 'text'
+    value: string
+    constructor(token: string) {
+        this.value = token
+    }
+}
+
+export class IntVariable {
```
cyanfish, line 19: Is all this code used?

src/commands.ts
```diff
@@ -0,0 +1,222 @@
...
+    }
+}
+
+type Variable = TextVariable | IntVariable
+type VariableConstructor = new (token: string) => Variable
+const VARIABLE_TYPES: Record<string, VariableConstructor> = {
+    text: TextVariable,
+    int: IntVariable,
+}
+export function isText(v: Variable): v is TextVariable {
```
cyanfish, line 34: To verify (since I'm not familiar with this construction), this means that a true return value implies the type of v is TextVariable?

src/config.ts
```diff
@@ -0,0 +1,358 @@
+import moment from 'moment'
```
cyanfish, line 1: Maybe name this file something else since we have config.js in the other folder? Also add a file comment

cyanfish:
Looks fine overall, made some comments. Also the readme needs updating.

src/commands.ts
lakinwecker: Yes, it's called a 'type guard' in typescript and it allows the compiler to know that the variable is that type from then on out

cyanfish:
One other thing I forgot, tests failed without a chesster_slack_token.js file, probably should fix that and/or add a chesster_slack_token.js.example file + doc.

config/config.js
```diff
@@ -1,295 +1,292 @@
...
+        logging: false,
+        pool: {
+            max: 5,
+            min: 0,
+            idle: 10000,
+        },
     },
-    "storage": "./db/database.sqlite",
-    "watcherBaseURL": "https://lichess.org/api/stream/games-by-users",
+    storage: '',
```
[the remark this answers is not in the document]
lakinwecker: Because we use postgresql now and the connection paramaters are above

config/config.js
```diff
@@ -1,295 +1,292 @@
...
-    "links": {
-        "source": "https://github.com/endrawes0/Chesster"
+    links: {
+        source: 'https://github.com/endrawes0/Chesster',
     },
-    "welcome": {
-        "channel": "general"
+    heltour: {
+        token: heltour_token,
+        baseEndpoint: 'http://localhost:8000/api/',
```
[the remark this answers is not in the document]
lakinwecker: Good catch, thanks.

src/commands.ts
lakinwecker: IntVariable is, it's used in the availability commands.

src/config.ts
lakinwecker: I added the file comment.

I'm not sure I want to rename this one. I'd rather rename the config.js file. Maybe we name it chesster_config.js or something. Is it ok if I just file an issue about that, rather than fixing it right now?

src/commands.ts
lakinwecker: Fixed., Good suggestion.

src/commands.ts
lakinwecker: Were there other parts of this that you were concerned were unused?

src/config.ts
lakinwecker: The primary reason is that I think I want to make even more improvements at the same time, so dealing with the rename in isolation is unnecessary work.

src/chesster.ts
```diff
@@ -0,0 +1,334 @@
...
+        new winston.transports.Console({
+            format: winston.format.simple(),
+        })
+    )
+}
+
+const adminSlack = new slack.SlackBot(
+    'chesster',
+    configFile,
+    false,
```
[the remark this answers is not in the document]
lakinwecker: Done.

Issue #333, opened 2020-07-06
Improvement for Chesster automated response


Sesquipedalism:
To those who can fix this, chesster sends a message to the scheduling channels when a time is entered that is not in the acceptable window, or when you try to use syntax like "Monday 2100" to schedule for today when it's the first day of a new round.  This often confuses people, so I think we should update the message to be a bit more helpful.  Thoughts on my proposed changes?

Current message:
'@'player Hi! Sorry, that time you posted is not an acceptable time. We need all games to end by 23:00 GMT on Monday, and we believe if you start then, you won't be done then! Please try and find a better time, and if you cannot, please contact the moderators.

Proposed message:
'@'player Hi! Sorry, that is not an acceptable time.  It looks like if you start at that time, you won't be done by the end of the round!  If you are scheduling for today, I sometimes get confused - try using the date and time format 'yyyy-mm-dd hhmm'.  Otherwise, please try and find a better time, and if you cannot, please contact the moderators.

lakinwecker:
Fix deployed. Thanks @mirkojotic !

PR #334, opened 2020-07-07
changing of phrasing in late scheduling message

mirkojotic:
Addressed #333


lakinwecker:
I have no idea how to trigger a rebuild with travis via this PR.

PR #355, opened 2020-10-16
fixes #353

### LINKED ISSUE #353: Chesster warning message about games very close to deadline contain personal user information
When scheduling a game very close to the end of the round, chesster sends a warning message. That message seems to include the account name of the email address the player registered with. So, it can contain the real name of the player, instead of the slack name. The message should refer to the slack account name.
Scheduling

dombrno:
Hello,
I was not able to test this yet, but maybe for such a simple fix the PR could go through.


lakinwecker:
99% certain this is the correct fix, but I'll test it before I merge.

dombrno:
Sure, I would do the same ;)

PR #387, opened 2021-10-30
Fixes #357 Notify team channels of alt games

### LINKED ISSUE #357: Chesster doesn't post games for alternates in their team channel
The notification system mistakenly checks for the team from the player, and the alternate doesn't have a team, so it doesn't work. It should look up the pairing and figure out both teams from this.

mathgeek:


lakinwecker:
So, while I think this _will_ work, I'm not sure it's the right design. I think I'd prefer it if the `getTeamByPlayerName` would also return the appropriate team name for alternates.  

I'm more than willing to merge this and run with this for the next season, but thought I would mention it here.

mathgeek:
> So, while I think this _will_ work, I'm not sure it's the right design. I think I'd prefer it if the `getTeamByPlayerName` would also return the appropriate team name for alternates.
> 
> I'm more than willing to merge this and run with this for the next season, but thought I would mention it here.

I thought about maybe taking that approach after I'd already done this one, let me see if I can get that working today

src/league.ts
```diff
@@ -506,17 +506,19 @@ export class League {
...
         playerName = playerName.toLowerCase()
         const directTeamMapping =
-            this._playerLookup[playerName.toLowerCase()] &&
-            this._playerLookup[playerName.toLowerCase()].team
+            this._playerLookup[playerName] &&
+            this._playerLookup[playerName].team
         if (directTeamMapping) return directTeamMapping
         // Try to find the team by looking through the pairings for this
         // playername.  This will find alternates.
-        const teams: Team[] = this._teams.filter((t) => {
```
mathgeek, line 514: This didn't actually work I think - looks like Heltour only returns TeamMembers in the players list

src/league.ts
```diff
@@ -506,17 +506,19 @@ export class League {
...
-            const playerIds = t.players.map((p) => p.username.toLowerCase())
-            return playerIds.filter((n) => n === playerName).length > 0
-        })
-        if (teams.length > 0) {
-            return teams[0]
+        const possiblePairings = this.findPairing(playerName)
```
mathgeek, line 514: One thing I noticed - findPairing compares usernames using .includes() instead of lodash.isEqual(), feels like that should probably be changed?

mathgeek:
CI works!

src/league.ts
```diff
@@ -503,28 +504,33 @@ export class League {
...
 
     // -------------------------------------------------------------------------
     // Get the team for a given team name
     // -------------------------------------------------------------------------
-    getTeam(teamName: string) {
+    getTeam(teamName: string | undefined): Team | undefined {
+        if (!isDefined(teamName)) {
+            return undefined
+        }
         return this._teamLookup[teamName.toLowerCase()]
```
lakinwecker, line 534: `        return this._teamLookup[teamName?.toLowerCase()]`

won't this return undefined without the if?

src/league.ts
lakinwecker: Yeah, we should be comparing for equality there. Good catch.

src/league.ts
lakinwecker: I think we should consider updating Heltour to include alternates in the team list during the time periods where they are part of a team. However, cyanfish may disagree. So I'm fine with this as a fix.

src/league.ts
lakinwecker: Although if we do change that, then the signature for that function should probably be `Pairing | undefined`

lakinwecker: Let's open a new issue for this one and get the rest of this live once you update the checks.

lakinwecker (approved):

Issue #109, opened 2016-07-06
Don't notify people of their own events


lakinwecker:
If you subscribe to a team, there is no real point in getting a notification when you are the cause of the event.

PR #388, opened 2021-11-01
Game link orientation

mathgeek:


test/test_subscription.ts
```diff
@@ -0,0 +1,87 @@
...
+//                 [whiteNoListen.name, blackListen.name],
+//                 {
+//                     league: mockLeague,
+//                     details: mockDetails,
+//                     white: whiteNoListen,
+//                     black: blackListen,
+//                 }
+//             )
+//
+//             // TODO: Add asserts on formatSpy - asynchronous nature of event handler breaks spies
```
mathgeek, line 83: Played around with this for a while - spies always get triggered before the async event handler gets called. Open to any ideas for potentially finding a way around it.

src/commands/subscription.ts
```diff
@@ -423,19 +448,21 @@ export function getListeners(
...
         // user and teams. This is somewhat wasteful - but I'm not sure whether
         // this is the right design or if we should pass in users to this point.
-        const teamNames = _(sources)
-            .map((s) => _league.getTeamByPlayerName(s))
-            .filter(isDefined)
-            .map('name')
-            .map(_.toLower)
-            .value()
-        const possibleSources = _.concat(sources, teamNames)
+        const teamNamesWithContext: SourceWithContext[] = sources
```
mathgeek, line 459: This was my first crack at a data model for passing around the source context - could potentially be extended to add whether notification came from a team (which could help solve #109). This is definitely the biggest change here.

src/commands/subscription.ts
```diff
@@ -541,7 +597,7 @@ export async function tellMeWhenHandler(
 // -----------------------------------------------------------------------------
 export async function helpHandler(bot: SlackBot, message: CommandMessage) {
```
lakinwecker, line 598: Should we also have an explicit return type here that will help us catch if we don't return properly in the future?  Not necessary for this PR, but in general, I'd like to see us rely less on the type-inference in cases like this.

mathgeek: I'm all in favor of more explicit typing

lakinwecker: Great. Same here.

lakinwecker: I don't mind type inference in callbacks, where it's obvious, but in top-level functions and the like, I prefer explicit typing.

src/commands/subscription.ts
```diff
@@ -18,6 +18,17 @@ export const emitter = new ChessLeagueEmitter()
 type Context = any
```
lakinwecker, line 18: We should get rid of these. :(  Fully meant to do that back when I first wrong them.  we can turn on eslint no-explicit-any and fix from there.

lakinwecker: In this case, I think you should probably go ahead and get rid of this any for this change, to make sure we're probably using the context throughout.

src/commands/subscription.ts
mathgeek: I think I'll open a ticket to go through all these handlers later, we should figure out what we want to do with them. Right now these callbacks are (bot, message) => void so we should probably decide if that's the paradigm we want to have in mind with them or if we want to clean them all up and start bubbling up promises for error handling and such

src/commands/subscription.ts
```diff
@@ -131,52 +168,73 @@ function formatNoTeamResponse() {
     return "You can't use my-team or my-team-channel since you don't have a team right now."
 }
 
+export function emitEvent(eventName: string, _league: League, sources: [string, string], context: Context): void {
```
mathgeek, line 171: It might even just make sense to get rid of the emitter pattern entirely, only thing it’s really using now is the reference to the bot object

PR #398, opened 2022-04-05
Some more explicit error messages for game link validation

mathgeek:


mathgeek:
Guess I should take another pass at some point on how to block on test failures and how to handle changing tests

lakinwecker (approved):

PR #400, opened 2022-04-05
Time range should be hours not ms

mathgeek:
Watcher was not automatically alerting on incorrect time control because the current time was more than 2 ms off the scheduled start time - this should be 2 hours.


lakinwecker:
Also deployed.

PR #407, opened 2022-09-27
Fixes #357 For real this time - bad property check

### LINKED ISSUE #357: Chesster doesn't post games for alternates in their team channel

mathgeek:
These functions exist but were unused before my (relatively new) code to detect alts was added. white_team is the property on the response object from Heltour, but whiteTeam is how it should exist on the TeamPairing object

### DIFF
--- a/src/heltour.ts
+++ b/src/heltour.ts
@@ -149,10 +149,10 @@ export const PairingsDecoder: Decoder<Pairing[]> = oneOf(
 
 export type Pairing = IndividualPairing | TeamPairing
 export function isIndividualPairing(obj: Pairing): obj is IndividualPairing {
-    return !obj.hasOwnProperty('white_team')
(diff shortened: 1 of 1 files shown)


mathgeek:
This check was failing at https://github.com/Lichess4545/Chesster/blob/main/src/league.ts#L293 and therefore not set at https://github.com/Lichess4545/Chesster/blob/main/src/league.ts#L526

hopffgam (approved):

lakinwecker:
I'm confused by your comment about these changes.  The changes may very well work, but what you're describing to me is unclear.

lakinwecker:
Made a ticket for fixing the lack of consistency: https://github.com/Lichess4545/Chesster/issues/408

PR #418, opened 2025-03-25
Migrate to Slack's Events API

MrScribblesChess:
Migrating to Slack's Events API.

The problem this solves:
-Previously, chesster ran on Slack's RTM API
-But, the RTM API will be shut down on April 1, 2025
-So we're migrating chesster and adminSlack to the Events API

Details:
-Updated environment variables in code and README
-Updated chesster.start() to connect using the new API
-Updated various listeners and utils to be compatible with the change

QA:
-Tested for all three leagues for these items:
-Chesster listens in #-scheduling and updates scheduled times accordingly
-Chesster picks up on game starts and results in #-games
-Chesster responds appropriately to its normall array of commands (`source`, `rating`, summon mods etc)
-Chesster forwards messages 

Sticking points:
-`@rating` (without a specified username) crashes chesster if the sender of the message isn't in the db
-My local heltour was only sending one message at a time to be forwarded, so I don't know for sure if Chesster will forward bulk messages

TODO:
This was done in a fairly intense time crunch because I only found out about the RTM API deprecation in late February. Once things are settled down, I'd like to:
-Abstract chesster stuff in to utils functions, particularly `this.app.event...` and `this.message...` processing of messages to Chesster
-Write unit tests for the utils. Test the logic, not the API


package.json
```diff
@@ -26,6 +28,7 @@
         "q": "^1.5.1",
         "sequelize": "^5.21.5",
         "sequelize-cli": "5.5.0",
+        "sqlite3": "^5.1.7",
```
lakinwecker, line 31: We don't depend on this in prod, so probably best for us to remove this as well

src/commands/subscription.ts
```diff
@@ -293,12 +297,15 @@ function processTellCommand(
                 target: target.toLowerCase(),
             },
         })
-            .then(() =>
-                resolve(
+            .then(([subscription, created]) => {
```
lakinwecker, line 300: The subscription/created parameters here are not necessary, are they?

src/commands/subscription.ts
```diff
@@ -519,19 +526,22 @@ export async function tellMeWhenHandler(
     message: CommandMessage
 ) {
     const convo = await bot.startPrivateConversation([message.user])
+    bot.say({
```
lakinwecker, line 529: Let's remove this.

src/models.ts
```diff
@@ -95,18 +95,19 @@ export async function connect(config: ChessterConfig) {
...
         )
     }
+
+    // TODO Events Api: This path will obviously have to be changed in prod, this is my local path
     const sequelize = new Sequelize(
-        config.database.name,
-        config.database.username,
-        config.database.password,
-        { ...config.database, dialect: 'postgres' }
+        'sqlite:/Users/a/Documents/personalprojects/chessterstuff/Chesster/chesster.db.sqlite'
```
lakinwecker, line 101: Yes, let's remove this.

src/models.ts
```diff
@@ -95,18 +95,19 @@ export async function connect(config: ChessterConfig) {
...
     )
 
     try {
+        winston.info('[models.connect()] Attempting to connect to database...')
```
lakinwecker, line 105: Take this out, or change it to a `.debug`

src/models.ts
```diff
@@ -95,18 +95,19 @@ export async function connect(config: ChessterConfig) {
...
         await sequelize.authenticate()
+        winston.info('[models.connect()] Database connection successful')
```
lakinwecker, line 107: Here too

src/models.ts
```diff
@@ -95,18 +95,19 @@ export async function connect(config: ChessterConfig) {
...
         defineModels(sequelize)
     } catch (e) {
-        winston.error(`[models.connect()] Error connection to db: ${e}`)
+        winston.error(`[models.connect()] Error connecting to db: ${e}`)
```
lakinwecker, line 110: Here too

src/slack.ts
```diff
@@ -1,7 +1,15 @@
 // -----------------------------------------------------------------------------
 // Bot / Slack related helpers
 // -----------------------------------------------------------------------------
-import { RTMClient } from '@slack/rtm-api'
+import { App } from '@slack/bolt'
+
+// Load environment variables; don't delete this
+// This is particularly for Events API stuff
+import dotenv from 'dotenv'
```
lakinwecker, line 8: THis shouldn't be in the live / prod version, if it is, it should be in a different file, like main.

src/slack.ts
```diff
@@ -604,79 +628,174 @@ export class SlackBot {
         this.config = config.ChessterConfigDecoder.decodeJSON(
             JSON.stringify(require(this.configFile))
         )
-        this.token = this.config.slackTokens[this.slackName]
+
+        // Confusing naming; chesster is the name of adminSlack; lichess4545 is the name of actual chesster
+        // @ts-expect-error it thinks token can be undefined here but it's incorrect
+        this.token =
+            this.slackName === 'lichess4545'
```
lakinwecker, line 635: This stuff should be moved to main and pass along to this class, rather than parsed here.

src/slack.ts
```diff
@@ -604,79 +628,174 @@ export class SlackBot {
...
+                : "It won't work without this"
 
         if (!this.token) {
             const error = `Failed to load token for ${this.slackName} from ${this.configFile}`
             this.log.error(error)
             throw new Error(error)
         }
+
+        // Confusing naming conventions; chesster is the name of adminSlack; lichess4545 is the name of actual chesster
+        const signingSecret =
```
lakinwecker, line 648: Same here

src/slack.ts
```diff
@@ -604,79 +628,174 @@ export class SlackBot {
...
+            this.slackName === 'lichess4545'
+                ? process.env.SLACK_SIGNING_SECRET // Chesster
+                : process.env.ADMIN_SLACK_SIGNING_SECRET // AdminSlack
+
+        // Confusing naming conventions; chesster is the name of adminSlack; lichess4545 is the name of actual chesster
+        const appToken =
+            this.slackName === 'lichess4545'
```
lakinwecker, line 655: And here

src/slack.ts
```diff
@@ -604,79 +628,174 @@ export class SlackBot {
...
+        // chesster is the name of adminSlack; lichess4545 is the name of actual chesster
+        const botToken =
+            this.slackName === 'lichess4545'
+                ? process.env.SLACK_APP_BOT_TOKEN // Chesster
+                : process.env.ADMIN_SLACK_BOT_TOKEN // adminSlack
+
+        this.web = new WebClient(botToken)
+
+        // TODO maybe these tokens should go in config or something?
+        this.app = new App({
```
lakinwecker, line 682: Yes, the tokens should go in the config.

src/slack.ts
```diff
@@ -899,7 +1017,22 @@ ${usernames.join(', ')}`
...
-        return this.web.chat.postMessage(options)
+        // Ensure attachments is always an array as required by Bolt's client
+        if (!options.attachments) {
+            options.attachments = []
+        }
+
+        // @ts-ignore
+        return this.app.client.chat.postMessage({
+            ...options,
+            attachments: options.attachments ? options.attachments : [],
```
lakinwecker, line 1028: This and the lines above are redundant

src/slack.ts
```diff
@@ -899,7 +1017,22 @@ ${usernames.join(', ')}`
             )
         }
         options.as_user = true
-        return this.web.chat.postMessage(options)
+        // Ensure attachments is always an array as required by Bolt's client
+        if (!options.attachments) {
+            options.attachments = []
```
lakinwecker, line 1022: This is redundant below

src/slack.ts
```diff
@@ -899,7 +1017,22 @@ ${usernames.join(', ')}`
...
+        }
+
+        // @ts-ignore
+        return this.app.client.chat.postMessage({
+            ...options,
+            attachments: options.attachments ? options.attachments : [],
+            reply_broadcast: options.reply_broadcast
+                ? options.reply_broadcast
```
lakinwecker, line 1030: This can be shortcuted with `!!options.reply_broadcast`

src/slack.ts
```diff
@@ -962,18 +1108,42 @@ ${usernames.join(', ')}`
                     listener.type === 'command' &&
                     allowedTypes.indexOf('command') !== -1
                 ) {
-                    _.map(listener.middleware, (m) =>
-                        m(this, { member, ...message })
+                    // IMPORTANT FIX: Log before callback
+                    this.log.info(
```
lakinwecker, line 1112: Why is logging before the callback necessary?

src/slack.ts
```diff
@@ -962,18 +1108,42 @@ ${usernames.join(', ')}`
...
                 } else if (
                     listener.type === 'league_command' &&
                     allowedTypes.indexOf('league_command') !== -1
                 ) {
                     if (_league && member) {
-                        // Typescript should have been able know that they are set appropriately
-                        // here.
-                        _.map(listener.middleware, (m) => m(this, message))
+                        // SAME FIX HERE
+                        this.log.info(
```
lakinwecker, line 1136: This fix makes no sense, same as before, do you know why?

src/slack.ts
```diff
@@ -1059,24 +1251,133 @@ ${usernames.join(', ')}`
...
+                    isPingModerator: false,
+                }
+
+                // Loop through ALL command listeners to find matches
+                let matched = false
+
+                // Find command listeners first
+                for (const listener of this.listeners) {
+                    if (
+                        listener.type === 'command' &&
```
lakinwecker, line 1327: Minor nit: if you invert the condition in this if, and you use `continue` instead, you can un-indent the following code and it makes things easier to understand

src/slack.ts
```diff
@@ -1086,8 +1387,11 @@ ${usernames.join(', ')}`
     }
 
     on(options: OnOptions) {
-        return this.rtm.on(options.event, (event) =>
-            options.callback(this, event)
-        )
+        if (options.event === 'member_joined_channel') {
+            this.app.event('member_joined_channel', async ({ event }) => {
+                // @ts-ignore hope I don't regret this lol
```
lakinwecker, line 1392: Yeah, I hope we don't too

utils/SlackEventsAPIUtils/chessterUtils.ts
```diff
@@ -0,0 +1,218 @@
+// -----------------------------------------------------------------------------
```
lakinwecker, line 1: Is this actually used aside from the tests? If not, let's remove it.

test/test_config.ts
```diff
@@ -127,8 +127,8 @@ describe('config types', function () {
                     channel: 'lonewolf-games',
                     channelId: 'C0SD3SCAH',
                     clock: {
-                        initial: 30,
-                        increment: 30,
+                        initial: 1,
```
lakinwecker, line 130: Yeah, all of these changes should be ignored. this config is setup to test on the live slack

lakinwecker: You can feel free to make and commit another config that you use for your slack, if you'd like

lakinwecker (changes requested):
Some small changes to consider.

package.json
```diff
@@ -15,6 +16,7 @@
         "codeclimate-test-reporter": "^0.5.1",
         "dotenv": "^8.2.0",
         "ee-ts": "^1.0.1",
+        "f": "^1.4.0",
```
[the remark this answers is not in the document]
lakinwecker: It doesn't look like we use it, we should remove it

utils/SlackEventsAPIUtils/chessterUtils.ts
lakinwecker: We can put it back in with a PR that causes it to be used, this sort of refactoring and tests is quite a good step towards what we want, but I don't like having unused code in master

PR #425, opened 2026-03-22
Blindfold league addition to chesster config

linuxexplorer:
Added the relevant blindfold league details to chesster config. Upon
chesster restart, scheduling messages in blindfold-scheduling should be
picked up by chesster once chesster bot is added to the channel


lakinwecker (approved):
LGTM

